PR #16, opened 2018-11-08
Add a variable to allow for customization of the directory from which tests are run

jsalzbergedu:
Hello,

So for my use case, I needed the tests to be run from a directory that is arguably the root directory,
but also arguably not. The project root that lsp-java finds is where the git repo starts, and the real project root is where the pom.xml is, one directory deeper. (I have this layout: `project-root-as-determined-by-lsp-java/project-root-as-I-see-it/test/blah/blah/blah/SomethingTest.java`. I cannot change the directory structure. However, by adding this variable, I am able to accommodate my usecase, which is admittedly an edge-case.

Thank you!

### DIFF
--- a/dap-java.el
+++ b/dap-java.el
@@ -60,6 +60,11 @@ If the port is taken, DAP will try the next port."
   :group 'dap-java
   :type 'number)
 
+(defvar-local dap-java-test-run-directory nil
+  "The directory from which dap-java will run a test.")
+
+(put 'dap-java-test-run-directory 'safe-local-variable #'stringp)
+
 (defun dap-java--select-main-class ()
   "Select main class from the current workspace."
   (let* ((main-classes (lsp-send-execute-command "vscode.java.resolveMainClass"))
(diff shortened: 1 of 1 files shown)


yyoncho:
> I needed the tests to be run from a directory that is arguably the root directory,
but also arguably not.

It is a bug discussed here: https://github.com/yyoncho/dap-mode/issues/10#issuecomment-424842271

1. Now there are the following methods in `lsp-java` - `lsp-java--find-project-uri` and `lsp-java-update-project-uris` they would fix the wrong root calculation. I haven't had a chance to test how reliable are they and whether it would be better just to call  `che.jdt.ls.extension.mavenProjects` for the current folder uri.

2. The other question is whether to allow customization of the properties in this manner: I believe that allowing:

``` emacs-lisp
(dap-java-run-test-method  (list :cwd "folder you want"))
```
would work better in the context of the current design since it will allow customizing all of the properties.

There is one piece that is missing from `dap-mode` - ability to edit the run spec like you can do in vscode. I haven't come up to a UI design how to do that. Here it is what vscode has: https://cdn-images-1.medium.com/max/1600/1*cSN9sQJgwMXolVTpmdEAkA.jpeg .

jsalzbergedu:
Hmm. What do you think of using .dir-locals.el for that? If not, would a simple file with an alist do?

yyoncho:
> Hmm. What do you think of using .dir-locals.el for that? If not, would a simple file with an alist do?

I currently have a list of alists(list of all configuration that were run) - the point is how to present that to the user. Also, there is completion support in that file, thus .dir-locals.el wont fit easily I think. Show be something like: select template and edit it before running.

yyoncho:
https://github.com/yyoncho/dap-mode/commit/6d308df22a6104da78967989cf09cda07cc597c4 

I have added a command `dap-debug-edit-template` which could be used to generate a run command which you could customize before executing.

PR #21, opened 2018-12-01
Fix an error with repl completions in python

dvzubarev:
There was an error after selecting a candidate:
```company-call-backend-raw: Company: backend company-dap-ui-repl error
"Wrong type argument: char-or-string-p, nil" with args (post-completion long_name)```


yyoncho:
Thanks!

PR #22, opened 2018-12-01
Add support for latest ptvsd

dvzubarev:


dap-python.el
```diff
@@ -37,8 +37,9 @@ If the port is taken, DAP will try the next port."
 
 (defun dap-python--populate-start-file-args (conf)
   "Populate CONF with the required arguments."
-  (let ((debug-port (dap--find-available-port "localhost" dap-python-default-debug-port)))
-    (compile (format "python -m ptvsd --port %s %s" debug-port buffer-file-name))
+  (let* ((host "localhost")
+         (debug-port (dap--find-available-port host dap-python-default-debug-port)))
+    (compile (format "python -m ptvsd --wait --host %s --port %s %s" host debug-port buffer-file-name))
```
yyoncho: IMO we should either remove the `:host` from the debug template or use it here.

dvzubarev: I think this provider is meant to run the debugger on localhost. So I removed :host from the template. If someone wants to run ptvsd on a remote host and then attach to it, he should create another provider for this, I suppose.

PR #24, opened 2018-12-27
Two small code fixes

danielmartin:
This PR does the following:

- Remove an unused `workspace` let binding.
- Require `cl` in `dap-mode`, as it uses `first` and `rest` and a user may receive a `Symbol's function definition is void: first` error message when running `dap-debug` otherwise.


yyoncho:
Thanks

PR #26, opened 2018-12-31
Remove duplicate link to DAP page in README

iNecas:
Also, s/It similar/It's similar/ and add the note about requiring lsp
mode to separate line.

Every contribution counts :)


yyoncho:
Thanks!

PR #28, opened 2019-01-05
Add a variable to get the debug file path 

xhcoding:
Hello
It is not work to get the path of debugging file with `buffer-file-name` in c/c++ mode.So I added a variable to let the user customize the function.

### DIFF
--- a/dap-lldb.el
+++ b/dap-lldb.el
@@ -34,13 +34,20 @@
   :group 'dap-lldb
   :type '(repeat string))
 
+(defcustom dap-lldb-debugged-program-function 'buffer-file-name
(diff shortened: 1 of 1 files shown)


yyoncho:
I will merge it but it seems like this approach should be revisited in the future. VScode has the ability to edit the Debug template but I haven't figured out how to make it work with emacs.

yyoncho:
@xhcoding 

https://github.com/yyoncho/dap-mode/commit/6d308df22a6104da78967989cf09cda07cc597c4 

I have added a command `dap-debug-edit-template` which could be used to generate a run command which you could customize before executing which could be useful for your usecase.

PR #29, opened 2019-01-06
Fix tests

danielmartin:
This PR fixes the integration tests so that they are runnable again.

(They are still a bit flaky, we should probably mock the server interface with https://github.com/Microsoft/vscode-mock-debug at some point.)


yyoncho:
Thanks for doing that! But really there are tasks that are much more fun(like implementing Watches/exception breakpoints).

PR #39, opened 2019-02-02
Add custom python executable

nasyxx:
In some cases, people's python executable is not `python`; it can be `python3`, `pyenv exec python` or others.


yyoncho:
Thanks for the PR. In general, I have to change the adapters to allow doing `dap-debug-edit-template` and to allow customizing all of the properties without the need to add defcustom for each of  them.

PR #40, opened 2019-02-05
Fix ptvsd format arguments

Zarig:
A recent commit #39 broke the Python dap-mode configuration. 

```
python -m ptvsd --wait --host python --port localhost 32000
Error: invalid --port <port>: invalid literal for int() with base 10: 'localhost'
```

I believe it is this line which causes the error: https://github.com/yyoncho/dap-mode/blob/46e1e9d5b2dd94d7044f8fac912877b6c2f6728d/dap-python.el#L49

There is an additional `dap-python-executable` in the arguments, which causes it.

### DIFF
--- a/dap-python.el
+++ b/dap-python.el
@@ -46,7 +46,7 @@ If the port is taken, DAP will try the next port."
   (let* ((host "localhost")
          (debug-port (dap--find-available-port host dap-python-default-debug-port))
          (python-executable dap-python-executable))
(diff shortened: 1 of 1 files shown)


yyoncho:
@Zarig we should either delete dap-python-executable and use python-executable or we should remove python-executable and keep dap-python-executable.

(cc @nasyxx )

nasyxx:
So sorry because of my negligence.  

In my local area, I am using `python-executable`.

PR #43, opened 2019-02-12
Tomterl fix type

tomterl:
Hmm -- don't know why the 23.Jan commits are included here...

Fix the `:type` error (`php`, not `node`)


yyoncho:
Thanks!

Issue #44, opened 2019-02-14
Symbol’s function definition is void: first


wenxin-wang:
When I tried to use `dap-lldb` for debugging c file, after running `M-x dap-debug` I got `Symbol’s function definition is void: first`.

lsp-mode and dap-mode are installed as:

```elisp
(use-package lsp-mode
  :commands lsp
  :init
  (setq lsp-prefer-flymake nil
	lsp-session-file (expand-file-name ".lsp-session-v1" my:cache-dir)))

(use-package lsp-ui
  :commands lsp-ui-mode
  :init
  (add-hook 'lsp-mode-hook #'lsp-ui-mode))

(use-package dap-mode
  :config
  (dap-mode 1)
  (dap-ui-mode 1))

(require 'dap-lldb)
(require 'dap-python)
```

If `debug-on-error` is turned on, the debugger shows that

```
Debugger entered--Lisp error: (void-function first)
  first(("LLDB Run Configuration" :type "lldb" :cwd nil :request "launch" :program nil :name "LLDB::Run"))
  #f(compiled-function (it) #<bytecode 0x16c48b5>)(("LLDB Run Configuration" :type "lldb" :cwd nil :request "launch" :program nil :name "LLDB::Run"))
  mapcar(#f(compiled-function (it) #<bytecode 0x16c48b5>) (("LLDB Run Configuration" :type "lldb" :cwd nil :request "launch" :program nil :name "LLDB::Run")))
  dap--completing-read("Select configuration template: " (("LLDB Run Configuration" :type "lldb" :cwd nil :request "launch" :program nil :name "LLDB::Run")) first nil t)
  byte-code("\301\302\303\010\304\305\306%A!C\207" [dap--debug-template-configurations copy-tree dap--completing-read "Select configuration template: " first nil t] 7)
  call-interactively(dap-debug record nil)
  command-execute(dap-debug record)
[...]

PR #45, opened 2019-02-14
dap-mode.el: use cl-lib instead of cl

wenxin-wang:
In order to solve #44 and (some problem of) #44 , https://github.com/yyoncho/dap-mode/commit/ebb34c394067c81e29a36dc71452f326319a782d introduces `(require 'cl)` outside of `eval-when-compile`, which causes complain from linter.

[Emacs recommands](https://www.gnu.org/software/emacs/manual/html_mono/cl.html#Organization) `cl-lib` instead of `cl` for new packages. However, functions provided by `cl-lib` are prefixed with 'cl-' (which provides some kind of namespace), so switching to `cl-lib` requires prepending this prefix to functions from `cl`.

After changing `cl` to `cl-lib` (restart emacs, `features` shows that `cl` is not loaded), all unknown symbols ( `first`, `second`, `rest`, `for`, `let*`) reported by `byte-recompile-directory` are changed to their `cl-lib` counterpart (`cl-first`, `cl-second`, `cl-rest`, `cl-for`, `cl-let*`).

### DIFF
--- a/dap-java.el
+++ b/dap-java.el
@@ -76,7 +76,7 @@ If the port is taken, DAP will try the next port."
     (cond
      ((= main-classes-count 0) (error "Unable to find main class.
 Please check whether the server is configured propertly"))
-     ((= main-classes-count 1) (first main-classes))
+     ((= main-classes-count 1) (cl-first main-classes))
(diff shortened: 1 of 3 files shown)


yyoncho:
Thanks for fixing this!

PR #49, opened 2019-02-24
Add support for setting and removing breakpoints using the mouse

danielmartin:
A click on the left margin or fringe will add a DAP breakpoint if
`dap-mode` minor mode is also enabled.

In the future, when enabling/disabling breakpoints is implemented,
this could toggle between creating, and then enabling/disabling
breakpoints, with another mouse event (ie, right click?) reserved for
removing a breakpoint.


yyoncho:
Looks good! Later we may implement mouse hover + toolbar with icons like realgud/gud.

PR #50, opened 2019-02-27
Fix dap-java compile and attach

dakra:
I was trying out `dap-mode` with java
and simply run `dap-debug` with compile and attach.


yyoncho:
Thank you.

PR #59, opened 2019-03-22
Don't call lsp--uri-to-path since pathFormat is path

kiennq:
Since `dap--initialize-message` is using `pathFormat` with `path` instead of `uri`, parser side should stop call lsp--uri-to-path too. Else we will have a throw.
``` elisp
(list :command "initialize"
        :arguments (list :clientID "vscode"
                         :clientName "Visual Studio Code"
                         :adapterID adapter-id
                         :pathFormat "path"
                         :linesStartAt1 t
                         :columnsStartAt1 t
                         :supportsVariableType t
                         :supportsVariablePaging t
                         :supportsRunInTerminalRequest t
                         :locale "en-us")
        :type "request")
```

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -614,7 +614,7 @@ thread exection but the server will log message."
       (-let (((&hash "line" line "column" column "name" name) stack-frame)
              (source (gethash "source" stack-frame)))
         (if source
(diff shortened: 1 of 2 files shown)


yyoncho:
Thank you and congrats for your first `dap-mode` PR! :tada:

Issue #61, opened 2019-03-24
:cwd does not work for dap-python


f279801:
The compile command is currently being executed inside the dap-python--populate-start-file-args directly, which does not process the :cwd. We should only build up the program-to-start and let the dap-debug process it as usual.

yyoncho:
I guess the fix will be what you have suggested in https://github.com/emacs-lsp/dap-mode/issues/57 .

f279801:
Yes, if we're keeping the same idea of the current fix for that issue. It'll fix both.

yyoncho:
Ok, I will assign both issues to you. Thank you.

Issue #57, opened 2019-03-18
dap-debug-edit-template seems broken with python


f279801:
With dap-debug-edit-template, my understanding is that we adapt the configuration, evaluate the new value then the debug session will be started.
But since all the new value is modified in a new opened buffer, buffer-file-name seems no longer valid?

Or we have another proper WoW for dap-debug-edit-template?
/Tai

yyoncho:
Can you help with testing the following configuration for python:


```
(defun dap-python--populate-start-file-args (conf)
  "Populate CONF with the required arguments."
  (let* ((host "localhost")
         (debug-port (dap--find-available-port host dap-python-default-debug-port))
         (python-executable dap-python-executable)
         (python-args (or (plist-get conf :args) "")))

    (dap--put-if-absent conf :program-to-start
                        (format "%s -m ptvsd --wait --host %s --port %s %s %s"
                                python-executable host debug-port buffer-file-name python-args))
    (plist-put conf :debugServer debug-port)
    (plist-put conf :port debug-port)
    (plist-put conf :wait-for-port t)
    (plist-put conf :hostName host)
    (plist-put conf :host host)
    conf))
```

f279801:
Hi @yyoncho ,
It works but I have another idea as below, It'll refactor the way how edit-template works currently but to me it helps to provide a clearer wow for this stuff:
- Add one more parameter to python configuration, i.e: `:target nill`
- dap-python--populate-start-file-args will do similar stuff as your current fix. But before that it will also try to figure out the real target by checking for nil or empty :target and set it to buffer-file-name if it is the case. Otherwise, real target will be the configured one. After this it will build and append program-to-start into the configuration and return this back to the caller. By this we will not compile the command but signal the caller to do it for us by setting up the program-to-start.
- dap-debug-edit-template its turn will export the configuration to DAP Templates buffer. But instead of providing the dap-debug command, we just provide the dap-register-debug-template command. This ties what the function really does and its name. Also it provides the abilities for user to both overwritting the current configuration (by keeping the same name) and creating a new one (by change the name and/or other paramter) before evaluating. It is also really nice if dap-debug-edit-template would filter out the program-to-start from what it prints to DAP Template (it will be rebuilt later on when dap-debug is executed).
- After finishing edit the debug template, the normal dap-debug can be ran again with the same procedure (of course with new configuration/or more template).

How do you think about that proposal? If you are OK with that, I can also help to provide a PR in case you want.

/Tai

[...]

PR #64, opened 2019-04-01
#5761

f279801:
Hi @yyoncho,

Sorry for this late response, I've been back from a short vacation.
Below is the idea for #57 and #61. I don't work with Java so I'm not so sure if it will break the Java configuration or not. But I'll try to setup a workspace and test it tmr.

/f279

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -1021,22 +1021,6 @@ should be started after the :port argument is taken.
                    breakpoints)
           (lsp-workspace-set-metadata "Breakpoints" breakpoints))))))
 
-(defun dap--remove-debug-template (template-name &optional all)
-  "Remove template configurations with template-name out of the registered list.
-If all is t then remove all configurations with the given name. Otherwise, remove
-the first found one"
-  (let* ((lst dap--debug-template-configurations)
-	 (cf (cl-first lst))
-	 (break nil))
-    (while (and cf (not break))
-      (if (equal (cl-first cf) template-name)
-	  (progn
-	    (setq dap--debug-template-configurations
-		  (delete cf dap--debug-template-configurations))
-	    (if (not all) (setq break t))))
-      (setq lst (cdr lst))
-      (setq cf (cl-first lst)))))
-
 (defun dap-mode-line ()
   "Calculate DAP modeline."
   (when lsp-mode
@@ -1138,7 +1122,8 @@ arguments which contain the debug port to use for opening TCP connection."
   "Register configuration template CONFIGURATION-NAME.
 
 CONFIGURATION-SETTINGS - plist containing the preset settings for the configuration."
-  (dap--remove-debug-template configuration-name t)
+  (setq dap--debug-template-configurations
+	(delq (assoc configuration-name dap--debug-template-configurations) dap--debug-template-configurations))
   (add-to-list
    'dap--debug-template-configurations
    (cons configuration-name configuration-settings)))
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -1155,18 +1155,20 @@ CONFIGURATION-SETTINGS - plist containing the preset settings for the configurat
         (file-error (setq success t))))
     port))
 
-(defun dap--select-template ()
-  "Select the configuration to launch."
+(defun dap--select-template (&optional origin)
+  "Select the configuration to launch.
+If origin is t, return the original configuration without prepopulation"
   (let ((debug-args (-> (dap--completing-read "Select configuration template: "
                                               dap--debug-template-configurations
                                               'cl-first nil t)
                         cl-rest
                         copy-tree)))
-    (or (-some-> (plist-get debug-args :type)
-                 (gethash dap--debug-providers)
-                 (funcall debug-args))
-        (error "There is no debug provider for language %s"
-               (or (plist-get debug-args :type) "'Not specified'")))))
+    (if origin debug-args
+      (or (-some-> (plist-get debug-args :type)
+		   (gethash dap--debug-providers)
+		   (funcall debug-args))
+	  (error "There is no debug provider for language %s"
+		 (or (plist-get debug-args :type) "'Not specified'"))))))
 
 (defun dap-debug (debug-args)
   "Run debug configuration DEBUG-ARGS.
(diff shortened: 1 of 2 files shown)


dap-mode.el
```diff
@@ -1167,25 +1186,38 @@ after selecting configuration template."
...
+  (let ((debug-args (dap--select-template(not parg))))
+    (progn
+      (message "debug-args: %s" debug-args)
+      (with-current-buffer (or (get-buffer "*DAP Templates*")
+			       (with-current-buffer (get-buffer-create "*DAP Templates*")
+				 (emacs-lisp-mode)
+				 (current-buffer)))
+	  (goto-char (point-max))
+	  (insert
+	   (format "\n\n(dap-register-debug-template \"%s%s\"\n"
```
yyoncho: I will have to think more about that register call, it will slightly diverge from what VScode does. I think that something similar what suggested by @danielmartin . We need to decide also how we are going to persist the templaces, e. g. we may do that in `workspace-folder/.dap-templates` or we can support the vscode format.

dap-mode.el
```diff
@@ -1167,25 +1186,38 @@ after selecting configuration template."
...
+(defun dap-debug-edit-template (&optional parg debug-args)
+  "Edit registered template DEBUG-ARGS.
+When being invoked with prefix argument, poping up the prepopulated version of the template.
+Otherwise, return its original version. After registration, the new template can be used
+normally with dap-debug"
+
+  (interactive "P")
+  (let ((debug-args (dap--select-template(not parg))))
+    (progn
+      (message "debug-args: %s" debug-args)
```
yyoncho: debug leftover?

dap-mode.el
```diff
@@ -1122,6 +1138,7 @@ arguments which contain the debug port to use for opening TCP connection."
   "Register configuration template CONFIGURATION-NAME.
 
 CONFIGURATION-SETTINGS - plist containing the preset settings for the configuration."
+  (dap--remove-debug-template configuration-name t)
```
yyoncho: Why not smt like https://stackoverflow.com/questions/9812393/elisp-how-to-delete-an-element-from-an-association-list-with-string-key ?

dap-mode.el
```diff
@@ -1167,25 +1186,38 @@ after selecting configuration template."
...
+      (message "debug-args: %s" debug-args)
+      (with-current-buffer (or (get-buffer "*DAP Templates*")
+			       (with-current-buffer (get-buffer-create "*DAP Templates*")
+				 (emacs-lisp-mode)
+				 (current-buffer)))
+	  (goto-char (point-max))
+	  (insert
+	   (format "\n\n(dap-register-debug-template \"%s%s\"\n"
+		   (plist-get debug-args :name)
+		   (if parg " - Copy" "")))
```
yyoncho: probably this should generate unique name similar to what we do in `dap--calculate-unique-name` .

yyoncho: (not to be addressed as part of this PR).

yyoncho:
> But I'll try to setup a workspace and test it tmr.

Forgot to mention, I will cover Java testing.

dap-mode.el
```diff
@@ -1138,18 +1155,20 @@ CONFIGURATION-SETTINGS - plist containing the preset settings for the configurat
         (file-error (setq success t))))
     port))
 
-(defun dap--select-template ()
-  "Select the configuration to launch."
+(defun dap--select-template (&optional origin)
+  "Select the configuration to launch.
+If origin is t, return the original configuration without prepopulation"
```
danielmartin: Nit: We should capitalize "ORIGIN" in docs, following Emacs Lisp style.

dap-mode.el
```diff
@@ -1021,6 +1021,22 @@ should be started after the :port argument is taken.
                    breakpoints)
           (lsp-workspace-set-metadata "Breakpoints" breakpoints))))))
 
+(defun dap--remove-debug-template (template-name &optional all)
+  "Remove template configurations with template-name out of the registered list.
```
danielmartin: As mentioned, we should write function arguments in all caps in docs.

dap-mode.el
danielmartin: We could support the `launch.json` format, yes. Interoperability with VS Code is a good thing.

dap-mode.el
```diff
@@ -1138,18 +1155,20 @@ CONFIGURATION-SETTINGS - plist containing the preset settings for the configurat
         (file-error (setq success t))))
     port))
 
-(defun dap--select-template ()
-  "Select the configuration to launch."
+(defun dap--select-template (&optional origin)
+  "Select the configuration to launch.
```
danielmartin: Here we talk about "configuration" in the comment but the function name is about "templates". I've proposed some terminology around DAP concepts so that we can understand and model things things more clearly.

dap-mode.el
f279801: Yes, I completely agree with @yyoncho that we should differentiate between the template and the configuration. This is also the reason why I though about a new command for edit/create configuration based on given template registered by each debug provider. But as @yyoncho reminder, new command should be avoided.

@yyoncho @danielmartin how about re-using lsp-workspace-root, we can have dedicated file for each project by hashing the project path and concatenating it with project name. Then putting the file under our dap-configurations folder. Besides, I'm more preferred to put this under the same location with our .dap-breakpoints since we will not affect the workspace repo.
Similar modification should be done for dap-breakpoints as well. Breakpoints for a project should not be visible inside another project view.

dap-mode.el
f279801: sure, should be removed

dap-mode.el
f279801: It's just because I'm still new with lisp and don't know this one yet :).
Will be adapted in next PS. Anyway, it only remove one item a time, right? So we still need dap--remove-debug-template.

dap-mode.el
f279801: ok.

dap-mode.el
f279801: Done

dap-mode.el
f279801: Done

dap-mode.el
f279801: Yes, completely agree with you on this.
Actually, I did recommend a new command for this, i.e: dap-edit-debug-configuration in the original ticket but the concern was that it will complicate things a lot.
But maybe it is better to fix them all at the same time then we're good.

yyoncho: I think it is fine to split it into several commits, so let us work on making this PR a step into what we want to achieve.

dap-mode.el
yyoncho: > Besides, I'm more preferred to put this under the same location with our .dap-breakpoints since we will not affect the workspace repo.

We can make that configurable via variable, so we will support both options(we might calculate the storage path relatively to the project root which will work for full paths like /home/user/.storage/ and .storage). I do not like the idea about hashing since ideally the file should be human readable and the hashing is breaking this. 

> Similar modification should be done for dap-breakpoints as well. Breakpoints for a project should not be visible inside another project view.

You might have 2 project and one of them is referencing the other, in this case you want to hit the breakpoint from projectA even if you have started the debug process from projectB. This is  common for java projects.

dap-mode.el
yyoncho: Can we have more than one match assuming that we clear the matching record just before inserting a new one? We may consider changing this to hash-table which will remove the need from this code.

dap-mode.el
f279801: > We can make that configurable via variable, so we will support both options

Agree.
> the file should be human readable and the hashing is breaking this.

We still have the project name there, the hash is just only a short stuff to distinguish projects that have the same name but different location. Otherwise, we may need to have some other thing.
>You might have 2 project and one of them is referencing the other,

With two different breakpoint files, one can also references to files that belong to other project with a normal absolute path, right?

dap-mode.el
f279801: >Can we have more than one match

No for sure. It was stuck in my mind that we're living with a mixing configuration. Tkx for the good catch.
> changing this to hash-table

Yeah, this should be the way to go from beginning. Changing it now need much more adaption so I guess we can live with your suggestion.

dap-mode.el
f279801: Sure, that is the best approach.

dap-mode.el
yyoncho: Yeah, back then when I was writing this code I was still learning elisp.

dap-mode.el
yyoncho: > We still have the project name there, the hash is just only a short stuff to distinguish projects that have the same name but different location. Otherwise, we may need to have some other thing.

I was thinking about not having the project name at all. If you want per project configuration - use the dir local to redirect the configuration into custom file. By default, keep then in a simple list globally.  

> With two different breakpoint files, one can also references to files that belong to other project with a normal absolute path, right?

You can make it work somehow but this will require adding a method which is - "add breakpoint and ask for project to use" since you will have both projects active and dap-mode won't know which project the breakpoint is for.

yyoncho:
Thank you for the good work. We may continue discussion the next steps in different issue if you want to do more work regarding the templates or something else(e. g. the python debug adapter supports attaching to live processes, which is not implemented).

f279801:
Thank all for your comments so far.
@yyoncho sure, I'll find out which one I can work on.

PR #70, opened 2019-04-10
Add command to download vscode bits

kiennq:
Instead of having manually download the `VsCode` extensions, we should be able to automatically/effortlessly download those extensions right from `Emacs`.
This changes provided:
- `dap-utils-get-vscode-extension` to download `VsCode` extension to specified path.
- Update `dap-chrome`, `dap-firefox`, `dap-gdb-llb`, `dap-go`, `dap-php` and `dap-ruby` to leverage new utilities function.
- Create new `dap-go-setup` series of command to download `VsCode` extension.

### DIFF
--- a/README.org
+++ b/README.org
@@ -194,12 +194,14 @@
       #+END_SRC
       This will add the python related configuration to  ~dap-debug~.
 ** Ruby
-   - Download and extract [[https://marketplace.visualstudio.com/items?itemName=rebornix.Ruby][VSCode Ruby Extension]] . Make sure that ~dap-ruby-debug-program~ is: ~("node" path-to-main-js)~ where ~node~ is either "node" if nodejs is on the path or path to nodejs and ~path-to-main-js~ is full path ~./out/debugger/main.js~ which is part of the downloaded VScode package.
-   - Follow the instructions on installing ~rdebug-ide~ from [[https://github.com/rubyide/vscode-ruby/wiki/1.-Debugger-Installation][Ruby Debug Installation]]
-   - Put in your emacs configuration.
-     #+BEGIN_SRC elisp
-       (require 'dap-ruby)
-     #+END_SRC
+    - Download and extract [[https://marketplace.visualstudio.com/items?itemName=rebornix.Ruby][VSCode Ruby Extension]]. You can do that either by:
+      - Calling ~dap-ruby-setup~, the extension will be downloaded and all your path will be automatically set up.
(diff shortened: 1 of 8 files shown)


kiennq:
> 1\. Download and install the plugin when a user selects a template for a particular language. E. g. when user select template XXX dap-mode will check whether the extension has been downloaded and in case it is not it will ask the user.

I will leave this one in other PR, just that I still don't know how should I approach this yet. Note that when `dap-xxx` is frist required and there's no vscode extension downloaded yet, there will be a message  notifies user can use `dap-xxx-setup` to download.

> 2\. Drop version param - I believe that marketplace has an API for finding the latest version. My concern is that now we will have to manually bump the version of the plugin which might not be that bad assuming that the extension might have changed (e. g. path is different).

I drop it and replace with `latest`. I don't know if marketplace has public API yet, probing around a little bit and somehow I found this working.
P.S: According to [this](https://visualstudio.uservoice.com/forums/331878-visual-studio-marketplace/suggestions/17623456-public-rest-api-for-visual-studio-marketplace), there seems to be no public API yet.

yyoncho:
> I will leave this one in other PR, just that I still don't know how should I approach this yet. 

Sure, no problem. 

> I drop it and replace with latest. 

That is good enough.

yyoncho:
> I will leave this one in other PR, just that I still don't know how should I approach this yet. 

Sure, no problem. 

> I drop it and replace with latest. 

That is good enough.

yssource:
> For easier of setting up vscode extension, you only need call dap-gdb-lldb-setup after requiring dap-gdb-lldb.

How to call `dap-gdb-lldb-setup` ?

yssource:
before
  (require 'dap-gdb-lldb)

I need to add

 (require 'dap-utils)


``` elisp
(defun yssource-programming/post-init-cc-mode ()
  (require 'dap-utils)
  (require 'dap-gdb-lldb)
  (spacemacs/set-leader-keys-for-major-mode 'c++-mode "db" nil)
  (spacemacs/dap-bind-keys-for-mode 'cc-mode)
  )
```



And now it works for me.
Thanks.

kiennq:
Ohh, I forgot to add `(require 'dap-utils)` in `dap-gdb-lldb`.

yyoncho:
@kiennq no worries, I will fix it.

yyoncho:
Fixed with https://github.com/emacs-lsp/dap-mode/commit/d03578ade1c74814cb1f5eb49d0547a09e529c48 .

PR #71, opened 2019-04-14
Only navigate to source files if we have a path

danielmartin:
According to the spec, "path" is an optional property of a Source, so
in some cases, stack frames may have a "source" but no "path"
inside (for example, if the stack frame is OS library code). We now
only try to open a source file if there is a path attached.

Previously, in those cases we tried to execute `(find-file nil)`, which 
aborted the program.

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -611,15 +611,18 @@ thread exection but the server will log message."
   "Make STACK-FRAME the active STACK-FRAME of DEBUG-SESSION."
   (let ((lsp--cur-workspace (dap--debug-session-workspace debug-session)))
     (when stack-frame
-      (-let (((&hash "line" line "column" column "name" name) stack-frame)
-             (source (gethash "source" stack-frame)))
-        (if source
-            (progn (find-file (gethash "path" source))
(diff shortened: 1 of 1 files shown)


dap-mode.el
```diff
@@ -611,15 +611,18 @@ thread exection but the server will log message."
...
-                   (setf (dap--debug-session-active-frame debug-session) stack-frame)
-
+      (-let* (((&hash "line" line "column" column "name" name) stack-frame)
+              (source (gethash "source" stack-frame))
+              (path (gethash "path" source)))
+        (setf (dap--debug-session-active-frame debug-session) stack-frame)
+        ;; If we have a source file with path attached, open it and
+        ;; position the point in the line/column referenced in the
+        ;; stack trace.
+        (if (and source path)
```
yyoncho: at this point, it is impossible to have source=nil since we have already executed `(path (gethash "path" source))`

yyoncho: I think that this line

> (path (gethash "path" source))

should be changed to 

> (path (and source (gethash "path" source)))

yyoncho:
Thank you!

PR #73, opened 2019-04-17
Name output and stderr buffers according to session name, reuse if they exist

lane-s:
When using dap-mode, I find that the *out* and *stderr* buffers accumulate quite
rapidly. This option uses the previously created buffers if they are found.

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -112,9 +112,6 @@ The hook will be called with the session file and the new set of breakpoint loca
 (defvar dap--debug-configuration ()
   "List of the previous configuration that have been executed.")
 
-(defvar dap--use-last-output-buffer t
-  "Whether to use the *out* and *stderr* buffers from the previous session.")
-
 (cl-defstruct dap--debug-session
   (name nil)
   ;; ‘last-id’ is the last JSON-RPC identifier used.
@@ -128,7 +125,7 @@ The hook will be called with the session file and the new set of breakpoint loca
   (response-handlers (make-hash-table :test 'eql) :read-only t)
   ;; DAP parser.
   (parser (make-dap--parser) :read-only t)
-  (output-buffer (if dap--use-last-output-buffer (get-buffer-create "*out*") (generate-new-buffer "*out*")))
+  (output-buffer nil)
   (thread-id nil)
   ;; reference to the workspace that holds the information about the lsp workspace.
   (workspace nil)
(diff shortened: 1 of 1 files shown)


yyoncho:
This PR cannot go as it is since you will have the output of all applications that you have started into a single buffer. If you want to reuse the output buffer you may add a hook to dap-session-created-hook and delete the output buffer on startup or use dap-terminated-hook to clean the current session's output buffer.

lane-s:
What about if the output buffers were named according to the session name like with the stderr buffer?

yyoncho:
> What about if the output buffers were named according to the session name like with the stderr buffer?

Make sense.

yyoncho:
@lane-s as a side note, the initial name could be calculated base on the binary.

lane-s:
@yyoncho Wouldn't the initial name be read from the template configuration file?

yyoncho:
@lane-s yes, now the name comes from the template but each function that populates the arguments could override it.

lane-s:
So is the session name not final until the `dap--session-init-resp-handler` is called?

yyoncho:
> So is the session name not final until the `dap--session-init-resp-handler` is called?

Until `dap-start-debugging`

lane-s:
@yyoncho So is my latest commit ok? It seems like if the session name is finalized at `dap-start-debugging`, then just changing `dap--create-session` to set `output-buffer` based on the session name should make it so that the output buffer is reused if the session name is the same, but otherwise a new buffer is created.

yyoncho:
I will have to test that manually(I am almost done for today). We should ensure that it is not possible to have 2 debug session logging into one output buffer. Also, it would be good to do not reuse the session buffer if there is some output from the previous session, or at least if we clear it.

lane-s:
I was thinking of just leaving the output from the previous run in the buffer. It might also be a good idea to add auto scrolling to the output buffer. I think I could do that too if you want.

There's still some problems with it though. It seems that the sessions are still in the list returned by `dap--get-sessions` even after the debugger terminates. If that's intended, then the best solution is probably just to delete session's output/stderr buffers when delete session is called. 

It also seems to me that if the session is still in the list, then you should be able to switch to it and delete, but currently this doesn't seem to be the case.

yyoncho:
> I was thinking of just leaving the output from the previous run in the buffer. It might also be a good idea to add auto scrolling to the output buffer. I think I could do that too if you want.

I think that there is emacs mode that could do that out of the box.

> There's still some problems with it though. It seems that the sessions are still in the list returned by dap--get-sessions even after the debugger terminates. If that's intended, then the best solution is probably just to delete session's output/stderr buffers when delete session is called.
> It also seems to me that if the session is still in the list, then you should be able to switch to it and delete, but currently this doesn't seem to be the case.

You could subscribe to dap-terminate-hook and clean up the terminated session(we may introduce method `dap-cleanup-terminated-sessions`  similar to `dap-delete-all-sessions`). Note that it is virtually impossible to support all configurations out of the box, but we should make as many customizations as possible via exposing API which dap-mode users could use to customize it.

We could create a set of recipes which could be used by the users to create `their` personal debug environment. I have created one: https://github.com/emacs-lsp/dap-mode/wiki/HowTo:-Display-debug-windows-on-session-startup .

yyoncho:

yyoncho:
> It might also be a good idea to add auto scrolling to the output buffer. I think I could do that too if you want.

Someone mentioned that this works, but I am not 100% sure. 

https://emacs.stackexchange.com/questions/41242/how-to-autoscroll-a-window-to-always-show-the-end-of-a-growing-buffer/41248#41248

I haven't tried it if it does not work a PR will be much appreciated.

Issue #51, opened 2019-03-03
Add support for node debugger


yyoncho:
https://github.com/Microsoft/vscode-node-debug2

PR #83, opened 2019-05-15
Add js debugger using VsCode built-in Node debugger

kiennq:
Issue #51

### DIFF
--- a/dap-node.el
+++ b/dap-node.el
@@ -1,0 +1,64 @@
+;;; dap-node.el --- Debug Adapter Protocol mode for Node      -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2019  Kien Nguyen
+
+;; Author: Kien Nguyen <kien.n.quang@gmail.com>
+;; Keywords: languages
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


yyoncho:
Thank you for your work on `dap-mode`!

When I was testing the adapter I saw that sometimes the adapter returns uri prefixed with `<node_internal>/file:` . I was debugging a simple file out.js with no project file or whatever. Have you seen something like that?

dap-node.el
```diff
@@ -0,0 +1,67 @@
+;;; dap-node.el --- Debug Adapter Protocol mode for Node      -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2019  Ivan Yonchovski
+
+;; Author: Ivan Yonchovski <yyoncho@gmail.com>
```
yyoncho: Change the author to you.

dap-node.el
```diff
@@ -0,0 +1,67 @@
...
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+;; URL: https://github.com/yyoncho/dap-mode
+;; Package-Requires: ((emacs "25.1") (dash "2.14.1") (lsp-mode "4.0"))
+;; Version: 0.2
+
+;;; Commentary:
+;; Adapter for https://github.com/nodeide/vscode-node
```
yyoncho: this url is 404.

yyoncho:
Ok. So I tested the file that used to fail and now it works fine. So, I will merge it after you fix the doc issues.

dap-node.el
kiennq: Fixed

dap-node.el
kiennq: Fixed

kiennq:
I've updated the doc, can you please check?

PR #105, opened 2019-07-14
[imp] check for python env, when searching for executable

mpanarin:
Similar to `spacemacs` `spacemacs/python-execute-file` https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Blang/python/funcs.el#L437
we should pick the appropriate python executable, taking python virtual env into concern.

I simply added a method, that is a copy of https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Blang/python/funcs.el#L116

### DIFF
--- a/dap-python.el
+++ b/dap-python.el
@@ -41,11 +41,33 @@ If the port is taken, DAP will try the next port."
   :risky t
   :type 'file)
 
+(defun dap-python--pyenv-executable-find (command)
+  "Find executable taking pyenv shims into account.
+If the executable is a system executable and not in the same path
+as the pyenv version then also return nil. This works around https://github.com/pyenv/pyenv-which-ext
+"
+  (if (executable-find "pyenv")
+      (progn
+        (let ((pyenv-string (shell-command-to-string (concat "pyenv which " command)))
+              (pyenv-version-names (split-string (string-trim (shell-command-to-string "pyenv version-name")) ":"))
+              (executable nil)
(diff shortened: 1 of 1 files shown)


mpanarin:
tests are red on Java :thinking:

Seems that master is red as well

yyoncho:
> tests are red on Java 
> 
> Seems that master is red as well

You could ignore the tests.

Unfortunately, I do not have the expertise to test this PR. I am CC-ing @seagle0128 if he wants to test, if not, I will merge it as it is.

seagle0128:
Sorry. I am not using `pyenv`. Perhaps @smallzhan is able to help on it.

mpanarin:
@seagle0128 Its more about using the `executable-find` than `pyenv`. Pyenv is good, but i don't use it either. `executable-find` will work perfectly, when you have virtualenvs configured in emacs. Pyenv is there in case you use it and have a bit different workflow from a regualar venv.

seagle0128:
Thank you, @mpanarin . I understand. Although it looks fine in my code reviews, I still think it's better have some test cases to cover all scenarios.

mpanarin:
I'd be glad to help with those, but my lisp knowledge is extremely limited :\

But I'll see what i can do

yyoncho:
I tested it without pyenv and it works fine(so there is no regression compared to the current state)so I am proceeding with merging this PR.

PR #114, opened 2019-08-23
Remove unnecessary config element

tomterl:
The server port isn't set this way.  The line is a remnant of the trial and error phase of setting up php debugging.


yyoncho:
Thank you!

PR #115, opened 2019-08-23
Add 'stop on entry' debugging template for php

tomterl:
Sometimes (mostly when debugging command-line tools), I don't want to/can set a breakpoint, but just want to enter the debugger from the start.  This resembles gebens default behaviour, IIRC.

### DIFF
--- a/dap-php.el
+++ b/dap-php.el
@@ -58,6 +58,15 @@
                                    :cwd nil
                                    :request "launch"
                                    :name "Php Debug"
+                                   :args '("--server=4711")
+                                   :sourceMaps t
+                                   ))
+(dap-register-debug-template "Php Stop On Entry"
+                             (list :type "php"
+                                   :cwd nil
+                                   :request "launch"
+                                   :name "Php SOE Debug"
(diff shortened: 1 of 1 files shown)


yyoncho:
Are you aware of `dap-debug-edit-template`? The alternative is to add the property to the existing configuration and let the users create a template and customize that value. I could merge that as it is but I am not quite sure whether creating a new template for each combination of values is a good idea. FTR as part of https://github.com/emacs-lsp/dap-mode/issues/77 we will implement the VScode way to create a debug configuration and it will make distinguishing configuration and template a bit more clear.

tomterl:
It seems I'm falling for the confusion regarding configuration vs. template, too :) 
So my way of adding this may be completely of the mark, and I'm totally fine leaving it (or the correct version of it) in my personal setup.  But I consider this particular case an essential configuration/template -- it should be present without further configuration IMVHO.

tomterl:
Thanks! That makes my life easier - I hope my perception wrt. to this convenience-feature is not completely wrong.

Issue #101, opened 2019-07-03
Support OCaml language


XVilka:
There are a few different implementations of OCaml LSP servers:

- https://github.com/freebroccolo/ocaml-language-server
- https://github.com/jaredly/reason-language-server
- https://github.com/bryphe/merlin-language-server

And one DAP implementation:

- https://github.com/hackwaly/ocamlearlybird

PR #119, opened 2019-08-29
filter output buffer, label by category, and add output buffer scrolling

### LINKED ISSUE #120: difficult to distinguish stdout in output buffer when debugging nodejs
When debugging a nodejs app, it's difficult to see stdout. 

Consider a simple program:
```nodejs
for (let i = 0; i < 10; i++) {
  console.log(i);
}
```

We get this output in the output buffer:

```
ClientRequest/initializedebugStarted/usr/local/bin/node --inspect-brk=44447 main.js 
Debugger listening on ws://127.0.0.1:44447/368ff515-5369-4159-900d-2a097e861ca4
For help, see: https://nodejs.org/en/docs/inspector
targetDebugProtocolVersiontargetCountDebugger attached.
ClientRequest/launchnodeVersionsetBreakpointsRequestClientRequest/setBreakpointsreport-start-up-timingsClientRequest/configurationDoneClientRequest/stackTracenextRequestClientRequest/next0
ClientRequest/stackTracenextRequestClientRequest/nextClientRequest/stackTracenextRequestClientRequest/nextClientRequest/stackTracenextRequestClientRequest/next1
ClientRequest/stackTracenextRequestClientRequest/nextClientRequest/stackTracenextRequestClientRequest/nextClientRequest/stackTracenextRequestClientRequest/next2
ClientRequest/stackTracenextRequestClientRequest/nextClientRequest/stackTracenextRequestClientRequest/nextClientRequest/stackTracenextRequestClientRequest/next3
ClientRequest/stackTracenextRequestClientRequest/nextClientRequest/stackTracenextRequestClientRequest/nextClientRequest/stackTracenextRequestClientRequest/next4
ClientRequest/stackTracenextRequestClientRequest/nextClientRequest/stackTracenextRequestClientRequest/nextClientRequest/stackTracenextRequestClientRequest/next5
ClientRequest/stackTracenextRequestClientRequest/nextClientRequest/stackTracenextRequestClientRequest/nextClientRequest/stackTracenextRequestClientRequest/next6
ClientRequest/stackTracenextRequestClientRequest/nextClientRequest/stackTracenextRequestClientRequest/nextClientRequest/stackTracenextRequestClientRequest/next7
ClientRequest/stackTracenextRequestClientRequest/nexttarget/notification/onScriptParsedtarget/notification/onPausednextRequestClientRequest/next8
nextRequestClientRequest/nextnextRequestClientRequest/nextnextRequestClientRequest/next9
nextRequestClientRequest/nextClientRequest/stackTraceClientRequest/stackTraceClientRequest/stackTraceClientRequest/stackTraceClientRequest/stackTracenextRequestClientRequest/nextClientRequest/stackTrace
```

artenator:
I am pretty new to this project, so feedback is greatly appreciated! Thanks 

closes #120 

## Reasoning behind this PR
- When debugging an application (nodejs in my case), it is difficult to distinguish stdout from other DAP output in the output buffer. See #120 
- The output buffer did not scroll when new output was printed to the buffer.

# What I added:
## Filtering output based on category of the DAP output
You can now use the custom `dap-output-buffer-filter` to filter the output based on the categories you want to see. I have set the default as "stdout", because I thought that the telemetry output such as "FullSessionStatistics/SourceMaps/Overrides" was not very helpful when debugging a program.

Result:
Consider this simple program
```nodejs
for (let i = 0; i < 10; i++) {
  console.log(i);
}
```

With `dap-output-buffer-filter` set to `'("stdout")`, the output buffer now produces this output:
```
0
1
2
3
4
5
6
7
8
9
```

## Labelling categories in output
You can use the custom `dap-label-output-buffer-category` to show the category that the output belongs to in the output buffer.
Now, the output buffer formats like this when the custom is non-nil:
```telemetry: ClientRequest/initialize
telemetry: debugStarted
console: /usr/local/bin/node --inspect-brk=5755 main.js 
stderr: Debugger listening on ws://127.0.0.1:5755/619ee71b-ada8-419c-932c-e64ba663392c
For help, see: https://nodejs.org/en/docs/inspector
telemetry: targetDebugProtocolVersion
telemetry: targetCount
stderr: Debugger attached.
telemetry: ClientRequest/launch
telemetry: nodeVersion
telemetry: setBreakpointsRequest
telemetry: ClientRequest/setBreakpoints
telemetry: report-start-up-timings
telemetry: ClientRequest/configurationDone
telemetry: ClientRequest/stackTrace
telemetry: FullSessionStatistics/SourceMaps/Overrides
telemetry: target/notification/onScriptParsed
telemetry: target/notification/onPaused
telemetry: ClientRequest/disconnect
telemetry: debugStopped
```

## Proper Scrolling of Output buffer
The output buffer was not properly scrolling because although the buffer's point was being set to `(point-max)`, the window's point was not. This was fixed by adding `(set-window-point win (point-max))` after inserting text to the output buffer.

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -719,9 +719,7 @@ thread exection but the server will log message."
     (pcase event-type
       ("output" (let* ((event-body (gethash "body" event))
                        (formatted-output (dap--output-buffer-format event-body)))
-                  (if dap-output-buffer-filter
-                      (when (member (gethash "category" event-body) dap-output-buffer-filter)
-                        (dap--print-to-output-buffer debug-session formatted-output))
+                  (when (or (not dap-output-buffer-filter) (member (gethash "category" event-body) dap-output-buffer-filter))
                     (dap--print-to-output-buffer debug-session formatted-output))))
       ("breakpoint" (-when-let* (((breakpoint &as &hash "id") (-some->> event
                                                                         (gethash "body")
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -711,8 +711,7 @@ thread exection but the server will log message."
   (with-current-buffer (dap--debug-session-output-buffer debug-session)
     (let ((win (get-buffer-window (current-buffer))))
       (goto-char (point-max))
-      (insert str)
-      (set-window-point win (point-max)))))
+      (insert str))))
 
 (defun dap--on-event (debug-session event)
   "Dispatch EVENT for DEBUG-SESSION."
@@ -791,6 +790,12 @@ thread exection but the server will log message."
                                 (message "Unable to find handler for %s." (pp parsed-msg)))))))
             (dap--parser-read parser msg)))))
 
+(defun dap--create-output-buffer (session-name)
+  "Creates an output buffer with with name SESSION-NAME."
+  (with-current-buffer (get-buffer-create (concat "*" session-name " out*"))
+    (set (make-local-variable 'window-point-insertion-type) t)
+    (current-buffer)))
+
 (defun dap--make-request (command &optional args)
   "Make request for COMMAND with arguments ARGS."
   (if args
@@ -850,7 +855,7 @@ ADAPTER-ID the id of the adapter."
                           :launch-args launch-args
                           :proc proc
                           :name session-name
-                          :output-buffer (get-buffer-create (concat "*" session-name " out*"))
+                          :output-buffer (dap--create-output-buffer session-name)
                           :workspace lsp--cur-workspace)))
     (set-process-sentinel proc
                           (lambda (_process exit-str)
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -43,6 +43,11 @@
   :group 'dap-mode
   :type 'boolean)
 
+(defcustom dap-debug-output-filter '("stdout")
+  "If non-nil, a list of output types to display in the debug output buffer."
+  :group 'dap-mode
+  :type 'list)
+
 (defcustom dap-auto-show-output t
   "If non-nil, the output buffer will be showed automatically."
   :group 'dap-mode
@@ -683,14 +688,36 @@ thread exection but the server will log message."
   (run-hook-with-args 'dap-terminated-hook debug-session)
   (dap--refresh-breakpoints))
 
+(defun dap--debug-output-buffer-format (category output)
+  "Formats a string suitable for printing to the output buffer using CATEGORY and OUTPUT."
+  (let ((message (format "%s: %s" category output)))
+    (if (string= (substring message -1) "\n")
+        message
+      (concat message "\n"))))
+
+(defun dap--print-to-debug-output-buffer (debug-session str)
+  "Insert content from STR into the debug output"
+  (with-current-buffer (dap--debug-session-output-buffer debug-session)
+    (let ((win (get-buffer-window (current-buffer))))
+      (goto-char (point-max))
+      (insert str)
+      (set-window-point win (point-max)))))
+
 (defun dap--on-event (debug-session event)
   "Dispatch EVENT for DEBUG-SESSION."
   (let ((event-type (gethash "event" event)))
     (pcase event-type
-      ("output" (with-current-buffer (dap--debug-session-output-buffer debug-session)
-                  (save-excursion
-                    (goto-char (point-max))
-                    (insert (gethash "output" (gethash "body" event))))))
+      ("output" (let* ((event-category (->> event
+                                            (gethash "body")
+                                            (gethash "category")))
+                       (event-output (->> event
+                                          (gethash "body")
+                                          (gethash "output")))
+                       (formatted-output (dap--debug-output-buffer-format event-category event-output)))
+                  (if dap-debug-output-filter
+                      (when (member event-category dap-debug-output-filter)
+                        (dap--print-to-debug-output-buffer debug-session formatted-output))
+                    (dap--print-to-debug-output-buffer debug-session formatted-output))))
       ("breakpoint" (-when-let* (((breakpoint &as &hash "id") (-some->> event
                                                                         (gethash "body")
                                                                         (gethash "breakpoint")))
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -43,7 +43,7 @@
   :group 'dap-mode
   :type 'boolean)
 
-(defcustom dap-output-buffer-filter '("stdout")
+(defcustom dap-output-buffer-filter '("stdout" "stderr")
   "If non-nil, a list of output types to display in the debug output buffer."
   :group 'dap-mode
   :type 'list)
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -706,11 +706,19 @@ thread exection but the server will log message."
       (dap--output-buffer-format-with-category (gethash "category" output-body) (gethash "output" output-body))
     (gethash "output" output-body)))
 
+(defun dap--insert-at-point-max (str)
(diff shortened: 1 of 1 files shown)


dap-mode.el
```diff
@@ -683,14 +693,37 @@ thread exection but the server will log message."
...
+(defun dap--output-buffer-format (output-body)
+  "Formats a string suitable for printing to the output buffer using an OUTPUT-BODY."
+  (if dap-label-output-buffer-category
+      (dap--output-buffer-format-with-category (gethash "category" output-body) (gethash "output" output-body))
+    (gethash "output" output-body)))
+
+(defun dap--print-to-output-buffer (debug-session str)
+  "Insert content from STR into the output buffer associated with DEBUG-SESSION."
+  (with-current-buffer (dap--debug-session-output-buffer debug-session)
+    (let ((win (get-buffer-window (current-buffer))))
```
yyoncho: I think that following the output might be achieved with https://stackoverflow.com/questions/12663061/emacs-auto-scrolling-log-buffer (e. g. when the user wants to follow the output). The problem with this implementation is that if you are in the debug console and you want to copy few lines at the beginning of the buffer and at the same time there is output event the focus will be stolen. We had similar issue in the past with `lsp-log` and there we went on implementing the following logic - if the user is in the end of the buffer - keep him in the end. At that time, I was unaware of the StackOverflow solution. What do you think?

dap-mode.el
```diff
@@ -683,14 +693,37 @@ thread exection but the server will log message."
...
   "Dispatch EVENT for DEBUG-SESSION."
   (let ((event-type (gethash "event" event)))
     (pcase event-type
-      ("output" (with-current-buffer (dap--debug-session-output-buffer debug-session)
-                  (save-excursion
-                    (goto-char (point-max))
-                    (insert (gethash "output" (gethash "body" event))))))
+      ("output" (let* ((event-body (gethash "body" event))
+                       (formatted-output (dap--output-buffer-format event-body)))
+                  (if dap-output-buffer-filter
```
yyoncho: nit: seems like you could join the `if` and `when` via or, eg `(when (or (not dap-ouput-buffer-filter) (member...`

dap-mode.el
```diff
@@ -43,6 +43,16 @@
   :group 'dap-mode
   :type 'boolean)
 
+(defcustom dap-output-buffer-filter '("stdout")
```
yyoncho: what about stderr?

dap-mode.el
artenator: Good point! 

I had originally had it condensed down to the solution you proposed, but I thought it was less understandable in terms of intent, and decided to go with readability over conciseness because of this.

Although, if that's OK with you, I'm happy to change it back!

dap-mode.el
artenator: Will add stderr as well!

dap-mode.el
artenator: Ahh, I understand the problem with this implementation now. 

I will look into this stackoverflow solution and see if it works!

Thanks for the feedback! I really appreciate it.

artenator: The solution from stackoverflow works as expected! The window point moves automatically if the point is at the end of the buffer, but does not move if the user moves the point somewhere else in the buffer.

artenator:
@yyoncho I have made the changes. Let me know what you think! Thanks 😎

dap-mode.el
```diff
@@ -683,14 +693,34 @@ thread exection but the server will log message."
...
+(defun dap--output-buffer-format (output-body)
+  "Formats a string suitable for printing to the output buffer using an OUTPUT-BODY."
+  (if dap-label-output-buffer-category
+      (dap--output-buffer-format-with-category (gethash "category" output-body) (gethash "output" output-body))
+    (gethash "output" output-body)))
+
+(defun dap--print-to-output-buffer (debug-session str)
+  "Insert content from STR into the output buffer associated with DEBUG-SESSION."
+  (with-current-buffer (dap--debug-session-output-buffer debug-session)
+    (let ((win (get-buffer-window (current-buffer))))
```
yyoncho: win is not used. Also, can you wrap it in save-excursion?

dap-mode.el
```diff
@@ -758,6 +788,12 @@ thread exection but the server will log message."
                                 (message "Unable to find handler for %s." (pp parsed-msg)))))))
             (dap--parser-read parser msg)))))
 
+(defun dap--create-output-buffer (session-name)
+  "Creates an output buffer with with name SESSION-NAME."
+  (with-current-buffer (get-buffer-create (concat "*" session-name " out*"))
+    (set (make-local-variable 'window-point-insertion-type) t)
```
yyoncho: having this "hardcoded" will still keep the other issue(stealing the focus). To solve it, I propose adding a minor mode (e. g. `dap-output-buffer-mode` and enabling it in the output buffer. Then, if the user wants to follow the output he/she could write 

```
(add-hook 'dap-output-buffer-mode-hook (lamnda () (make-local-variable 'window-point-insertion-type))
```

yyoncho: ... FTR we need that minor mode either way, since so the users could have custom keybindings in the output and so on.

dap-mode.el
yyoncho: I missed that comment - please ignore the comments from the last review(except the one for redundant win). I will test the solution on my side later today and proceed with merging if everything is ok.

dap-mode.el
artenator: wrapping this in `save-excursion` seems to break the scrolling functionality that was added from setting the `window-point-insertion-type`. I'm assuming this happens because when we set `save-excursion`, the point moves back to its original position after insertion, but I'm not 100% sure.

dap-mode.el
artenator: OK sounds good!

dap-mode.el
artenator: The solution I've implemented is to check whether the output buffer is focused or not, and wrap `save-excursion` if it is focused. Let me know if this is suitable

dap-mode.el
artenator: I'd be happy to implement a minor mode like this! Let me know if the updated solution is feasible for now, if not, I will work on this instead!

yyoncho:
Thank you for contributing to `dap-mode`! And congrats for your first PR! :tada:

I tested it and it works as expected. Some of the potential next steps:

1. We could introduce the minor mode(so the user could customize the behaviour in the output buffer).
2. We could use different font-lock for a category prefix in the log entries (to make it more beautiful).

Anyway, if you are interested in helping the project feel free to ping me or pick an issue in emacs-lsp organization.

artenator:
> Thank you for contributing to `dap-mode`! And congrats for your first PR! 🎉
> 
> I tested it and it works as expected. Some of the potential next steps:
> 
> 1. We could introduce the minor mode(so the user could customize the behaviour in the output buffer).
> 2. We could use different font-lock for a category prefix in the log entries (to make it more beautiful).
> 
> Anyway, if you are interested in helping the project feel free to ping me or pick an issue in emacs-lsp organization.

Thanks! It was a lot of fun and I learned a lot.

I look forward to contributing more. I hope to implement both 1 and 2 that you mentioned soon, but I am very interested in implementing #101 . Let me know what you think. Thanks!

yyoncho:
> but I am very interested in implementing #101 . Let me know what you think!

Pick whatever feature/issue you would enjoy the most. Just drop a note in the issue that you will be working on to avoid someone else picking it up. If you have a questions for it ping me here or in gitter.

PR #124, opened 2019-09-17
Fix breakpoints broken in multibyte character environment

iRi-E:
In multibyte character environment, the breakpoints can be unusable for two reasons:

1. If there are multibyte characters in source code, the breakpoints will be shifted upward (see screenshot below)
Screenshot from 2019-09-17 00-50-21
2. If file path includes multibyte characters (e.g. `~/あいう/src/main.rs`), the breakpoints cannot be read from .dap-breakpoints

I confirmed these problem are caused by insert-file-contents-literally() and can be resolved by using insert-file-contents() instead.


yyoncho:
Thank you and congrats for your first `dap-mode` PR! :tada:

PR #137, opened 2019-10-11
Document how to get a JUnit JAR to run Java tests

langston-barrett:


yyoncho:
Can you include in the docs that lsp-java downloads the runner when you call lsp-java-update-server?

langston-barrett:
@yyoncho Ah, I was not aware that it would! I'll mention that.

PR #145, opened 2019-10-31
Allow multiple debug sessions when using :program-to-start

iRi-E:
`compile()` allows only one subprocess to run simultaneously unless renaming the buffer name.
Using `compilation-start()` instead can run multiple subprocesses and use arbitrary major mode.

Also, we can now get more suitable buffer name and log messages like "Debug Adapter started at ..." rather than "Compilation started at ...".


yyoncho:
Thank you!

PR #146, opened 2019-11-02
Fix failures trying to delete the process dead already

iRi-E:
I found deleting session(s) may fail trying to delete the process dead already.

We should consider the follwing things:
- Response to `configurationDone` request may be processed after the process is terminated
- The process is not running in `failed` state as well as `terminated` state

The fix seems to work well, but I'm not sure whether `dap-session-changed-hook` should be run in `dap--send-configuration-done()` even if the process has been terminated.

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -160,7 +160,7 @@ The hook will be called with the session file and the new set of breakpoint loca
   (cursor-marker nil)
   ;; The session breakpoints;
   (session-breakpoints (make-hash-table :test 'equal) :read-only t)
-  ;; one of 'pending 'started 'terminated 'failed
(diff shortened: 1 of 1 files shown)


yyoncho:
Thank you for taking care!

dap-mode process management was inspired by lsp-mode process management one year ago but since then we have improved lsp-mode a lot but dap-mode still have some rough edges. For example,  there is no detection whether a particular adapter exists and if you start it, it will fail with random error.

> The fix seems to work well, but I'm not sure whether dap-session-changed-hook should be run in dap--send-configuration-done() even if the process has been terminated.

we run a hook(s) each time the GUI has to change its representation.

PR #152, opened 2019-11-07
Fix dap-debug() and dap-debug-edit-template() not working correctly in LISP code

iRi-E:
When using `dap-debug()` or `dap-debug-edit-template()` in LISP code, the argument `DEBUG-ARGS` is not correctly processesed.

- `dap-debug()`:
  The docstring says "If `DEBUG-ARGS` is not specified ...", but it cannot be omitted.
- `dap-debug-edit-template()`:
  Specifying `DEBUG-ARGS` has no effect.  Also, that unused argument causes a compile warning.

Note that the fix swaps the 1st and 2nd arguments of `dap-debug-edit-template()`.

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -1243,68 +1243,64 @@ CONFIGURATION-SETTINGS - plist containing the preset settings for the configurat
         (file-error (setq success t))))
     port))
 
-(defun dap--select-template (&optional origin)
-  "Select the configuration to launch.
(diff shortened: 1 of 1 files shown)


yyoncho:
> The docstring says "If DEBUG-ARGS is not specified ...", but it cannot be omitted.

When you call an interactive function from elisp code you could do `(call-interactively 'dap-debug)`

PR #159, opened 2019-11-19
Correct dap--get-path-for-frame to work in Windows

kiennq:
In `Windows`, absolute path is treated as url with `url-type` is name of driver.
For example `C:\example` will be parsed as url with `url-type` of `C`.
This cases the problem for `lsp--uri-to-path`.
We need to filter out `url-type` of one character long for it to work properly.

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -651,9 +651,12 @@ thread exection but the server will log message."
 
 (defun dap--get-path-for-frame (stack-frame)
   "Get file path for a STACK-FRAME."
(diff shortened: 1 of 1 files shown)


yyoncho:
This seems to be a hack - ATM I am considering changing lsp--uri-to-path to leave the uri as it is when it cannot handle the type instead of throwing an error. WDYT?

kiennq:
It would also be fine to me. Should I create new PR in lsp-mode instead?

PR #186, opened 2020-01-04
Derive dap-server-log-mode from compilation-mode for stacktrace highlighting and links

gwyker:
Deriving from compilation-mode in the server-log buffer adds syntax highlighting and links for error stacktraces


yyoncho:
Thank you!

PR #207, opened 2020-01-31
Render ansi colors in output buffers

kurnevsky:
Before:
screenshot-2020-01-31-16:55:49
After:
screenshot-2020-01-31-16:55:57

I also made these buffers read-only.

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -32,6 +32,7 @@
 (require 'dash)
 (require 'dap-overlays)
 (require 'cl-lib)
+(require 'ansi-color)
 
 (defcustom dap-breakpoints-file (expand-file-name (locate-user-emacs-file ".dap-breakpoints"))
   "Where to persist breakpoints"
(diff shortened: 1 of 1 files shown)


yyoncho:
Thank you!

dap-mode.el
```diff
@@ -739,16 +740,19 @@ thread exection but the server will log message."
...
   "Inserts STR at point-max of the buffer."
   (goto-char (point-max))
-  (insert str))
+  (insert (ansi-color-apply str)))
 
 (defun dap--print-to-output-buffer (debug-session str)
   "Insert content from STR into the output buffer associated with DEBUG-SESSION."
   (with-current-buffer (get-buffer-create (dap--debug-session-output-buffer debug-session))
+    (font-lock-mode t)
+    (setq-local buffer-read-only nil)
```
brotzeit: Why not use a let binding here ?

kurnevsky: Because by default it's not set, and we make it read-only with `(setq-local buffer-read-only t)` below. If we use let binding we will still need this line, so it doesn't make much sense.

PR #218, opened 2020-02-07
Add note about posframe dependency

kurnevsky:


yyoncho:
posframe requires emacs 26 so it is optional dependency.

PR #239, opened 2020-03-24
Removed java mentioning from breakpoint description

Flygrounder:
Description for `dap-breakpoint-add` says `Add java breakpoint at line`, which seems to be a mistake because it is not tied to Java language.


yyoncho:
Thank you!

PR #240, opened 2020-03-24
Fix test if posframe is installed

dakra:
When posframe is installed but not required yet
`(fboundp 'posframe-show)` returns `t` because it's autoloaded
but posframe-hide is not autoloaded (because it doesn't make sense)
and therefore `(fboundp 'posframe-show)` returns `nil`
which then wrongly complains that I need `posframe` installed.


yyoncho:
Thank you!

PR #249, opened 2020-04-17
[dap-dart] Add dart debug support

ericdallo:
The `Dart-Code` VSCode dart extension does not contain the debugger program on the marketplace zip, so this fetches the package by version from Github then transpile the [`dart_debug_entry.ts`](https://github.com/Dart-Code/Dart-Code/blob/master/src/debug/dart_debug_entry.ts) to make able to DAP run it.

dap-dart


dap-dart.el
```diff
@@ -0,0 +1,105 @@
...
+     (lsp-async-start-process
+      (lambda () (message "DAP Dart :: Setup done!"))
+      (lambda () (message "DAP Dart :: Error setting up dap-dart, check if `npm` is on $PATH"))
+      (f-join dap-dart-debugger-path "extension/node_modules/typescript/bin/tsc")
+      "--project" (f-join dap-dart-debugger-path "extension")))
+   (lambda () (message "DAP Dart :: Error setting up dap-dart, check if `npm` is on $PATH"))
+   "npm" "install" "--prefix" (f-join dap-dart-debugger-path "extension")
+   "--no-package-lock" "--silent" "--no-save"))
+
+(add-hook 'dap-utils-extension-downloaded-hook #'dap-dart--setup-extension nil t)
```
yyoncho, line 75: this should not be on top-level - it will be registered in the buffer you have performed the require lsp-dart call. I suggest altering dap-utils-github-extension-setup-function to allow passing on finished callback or wrapping the setup function into another function.

ericdallo: Done @yyoncho!
I didn't noticed that, thanks

dap-dart.el
```diff
@@ -0,0 +1,105 @@
+;;; dap-dart.el --- Debug Adapter Protocol mode for Dart language -*- lexical-binding: t; -*-
```
yyoncho, line 1: I can merge it it as it is but IMO this should go in lsp-dart project considering the following points:

1. It will be easier for the users to configure everything - fewer things to install if you are trying to make emacs dart IDE.
2. You will own it and it will be easier for you to change it to add new templates.
3. You will reference it either way from lsp-dart in order to implement debug lenses.

ericdallo: I see... I think we I can create a `lsp-dart-dap.el` on `lsp-dart`, what do you think?

yyoncho: It is up to you to decide for the name. Having the adapter in lsp-dart.el or keeping the current name dap-dart.el is fine too.

ericdallo: @yyoncho, I'll move tihis file to `lsp-dart`, but can we merge the `dap-utils` change?

yyoncho:
Thank you!

ericdallo:
Added this feature [here](https://github.com/emacs-lsp/lsp-dart/pull/5)

PR #258, opened 2020-04-30
Add support for async conf

ericdallo:
In `lsp-dart` we need to ask/setup the emulator for `Flutter` when user chooses to `dap-debug`, so this adds support for async debug.

Confirmed that works with `dap-debug`, `dap-debug-last`, `dap-debut-recent` and `dap-debug-restart`.


ericdallo:
c/c @yyoncho

PR #265, opened 2020-05-05
Add support for running TestNG tests

haicorner:
To keep this simple, `dap-java-use-testng` flag is added. When set to
a non-nil value, `dap-java--run-unit-test-command` builds the programm
list to run TestNG; default is nil to run JUnit and keep the existing
behaviour unchanged.

Additionally, the directory where TestNG generates reports can be
customized via `dap-java-testng-report-dir`

### DIFF
--- a/dap-java.el
+++ b/dap-java.el
@@ -39,6 +39,9 @@
                                  "%%%s%%"
                                "$%s"))
 
+;; Set to non-nil to use TestNG instead of the default JUnit
+(defvar dap-java-use-testng nil)
+
 (defcustom dap-java-java-command "java"
   "Path of the java executable."
   :group 'dap-java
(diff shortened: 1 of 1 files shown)


haicorner:
Not sure if you have plans to support TestNG or not, but this works for me (I work on some projects that use JUnit and some that use TestNG).
A better way than using the -use-testng flag would be to automatically detect whether the test class uses TestNG based on the import of org.testng.annotations.Test (or org.testng.annotation*). But I'm not an elisp developer, so that might come in the future if you think this is useful.

yyoncho:
Thank you for the PR!

> A better way than using the -use-testng flag would be to automatically detect whether the test class uses TestNG based on the import of org.testng.annotations.Test (or org.testng.annotation*). But I'm not an elisp developer, so that might come in the future if you think this is useful.

The full support for testng will come from https://github.com/emacs-lsp/lsp-java/blob/master/lsp-jt.el . ATM I have implemented the Junit4 support but I havent had the time to finish the work.

PR #277, opened 2020-05-12
Add auto configuration mode

ericdallo:
This adds an option to auto-configure `dap-mode` and `dap-ui-mode` to have a more "Emacs as IDE" behavior.

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -140,10 +140,27 @@ The hook will be called with the session file and the new set of breakpoint loca
   :group 'dap-mode
   :type '(plist))
 
-(defcustom dap-auto-configure t
-  "Whether to auto configure dap-mode and dap-ui-mode"
+(defcustom dap-auto-configure-features '(sessions locals breakpoints expressions controls tooltip)
+  "Windows to auto show on debugging when in dap-ui-auto-configure-mode."
   :group 'dap-mode
-  :type 'bool)
+  :type '(set (const sessions)
+              (const locals)
+              (const breakpoints)
+              (const expressions)
+              (const controls)
+              (const tooltip)))
+
+(defconst dap-feature->start-stop
+  '((sessions . (dap-ui-sessions . dap-ui--sessions-buffer))
+    (locals . (dap-ui-locals . dap-ui--locals-buffer))
+    (breakpoints . (dap-ui-breakpoints . dap-ui--breakpoints-buffer))
+    (expressions . (dap-ui-expressions . dap-ui--expressions-buffer))))
+
+(defconst dap-feature->mode
+  '((sessions . dap-ui-sessions-mode)
+    (breakpoints . dap-ui-breakpoints-mode)
+    (controls . (dap-ui-controls-mode . posframe))
+    (tooltip . dap-tooltip-mode)))
 
 (defvar dap--debug-configuration nil
   "List of the previous configuration that have been executed.")
@@ -1242,8 +1259,6 @@ before starting the debug process."
 
 (defun dap--after-initialize ()
   "After initialize handler."
-  (when dap-auto-configure
-    (dap-auto-configure-mode 1))
   (with-demoted-errors
       "Failed to load breakpoints for the current workspace with error: %S"
     (let ((breakpoints-file dap-breakpoints-file))
@@ -1502,6 +1517,7 @@ If the current session it will be terminated."
                        (-when-let (buffer (dap--debug-session-output-buffer debug-session))
                          (kill-buffer buffer))
                        (dap--refresh-breakpoints))))
+    (message "=>%s" cleanup-fn)
     (if (not (dap--session-running debug-session))
         (funcall cleanup-fn)
       (dap--send-message (dap--make-request "disconnect"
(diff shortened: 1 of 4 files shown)


dap-mode.el
```diff
@@ -1598,6 +1606,27 @@ point is set."
...
+  :group 'dap-mode
+  (cond
+   (dap-auto-configure-mode
+    (dap-ui-mode 1)
+    (when (require 'posframe nil t)
+      (dap-ui-controls-mode 1))
+    (add-hook 'dap-stopped-hook #'dap-ui-show-debug-windows)
+    (add-hook 'dap-terminated-hook #'dap-ui-hide-debug-windows))
+   (t
+    (dap-ui-mode nil)
```
kiennq, line 1625:
```suggestion
    (dap-ui-mode -1)
```

dap-mode.el
```diff
@@ -1598,6 +1606,27 @@ point is set."
...
+    (dap-ui-controls-mode nil)
```
kiennq, line 1626:
```suggestion
    (dap-ui-controls-mode -1)
```

dap-mode.el
```diff
@@ -1598,6 +1606,27 @@ point is set."
...
+  "Auto configure dap minor mode."
+  :init-value dap-auto-configure
+  :group 'dap-mode
+  (cond
+   (dap-auto-configure-mode
+    (dap-ui-mode 1)
+    (when (require 'posframe nil t)
+      (dap-ui-controls-mode 1))
+    (add-hook 'dap-stopped-hook #'dap-ui-show-debug-windows)
+    (add-hook 'dap-terminated-hook #'dap-ui-hide-debug-windows))
```
yyoncho, line 1623: I would put this in a separate mode so it could be turned off and on independently.

dap-mode.el
```diff
@@ -1236,6 +1242,8 @@ before starting the debug process."
 
 (defun dap--after-initialize ()
   "After initialize handler."
+  (when dap-auto-configure
```
yyoncho, line 1245: dap-auto-configure seems to be not needed and the dap-auto-configure-mode should be called after lsp-managed-mode has been enabled. So, generally, we might move calling auto-configuration mode in lsp-mode.

dap-ui.el
```diff
@@ -1109,5 +1109,27 @@ DEBUG-SESSION is the debug session triggering the event."
...
+(defun dap-ui--window-visible-p (window-name)
+  "Return whether WINDOW-NAME is visible."
+  (-> (-compose 'buffer-name 'window-buffer)
+      (-map (window-list))
+      (-contains? window-name)))
+
+(defun dap-ui-show-debug-windows (_session)
+  "Show available debug windows for SESSION."
+  (save-excursion
+    (unless (dap-ui--window-visible-p dap-ui--locals-buffer)
```
yyoncho, line 1121: most likely we would need to have a list of windows to show up(or list of functions to call). We have also log window, loaded sources, expressions. The idea is to have a generic way the users to specify what windows to pop-up.

dap-mode.el
```diff
@@ -1494,6 +1517,7 @@ If the current session it will be terminated."
                        (-when-let (buffer (dap--debug-session-output-buffer debug-session))
                          (kill-buffer buffer))
                        (dap--refresh-breakpoints))))
+    (message "=>%s" cleanup-fn)
```
yyoncho, line 1520: leftover?

ericdallo: sorry haha, removing it..

dap-mode.el
```diff
@@ -1598,6 +1621,49 @@ point is set."
...
+(defun dap--hide-feature-window (_session)
+  "Hide all debug windows when sessions are dead."
+  (unless (-filter 'dap--session-running (dap--get-sessions))
+    (seq-doseq (feature-start-stop dap-auto-configure-features)
+      (-when-let* ((feature-start-stop (alist-get feature-start-stop dap-feature->start-stop))
+                   (buffer-name (symbol-value (cdr feature-start-stop))))
+        (and (get-buffer buffer-name)
+             (kill-buffer buffer-name))))))
+
+(define-minor-mode dap-auto-configure-mode
```
yyoncho, line 1642: global?

dap-mode.el
```diff
@@ -1598,6 +1621,49 @@ point is set."
   (interactive)
   (dap-mode t))
 
+
+;; Auto configure
+
+(defun dap--show-feature-window (_session)
+  "Show auto confiured feature windows."
```
kiennq, line 1628:
```suggestion
  "Show auto configured feature windows."
```

dap-mode.el
kiennq, line 1642:
```suggestion
;;;###autoload
(define-minor-mode dap-auto-configure-mode
```

dap-mode.el
```diff
@@ -1598,6 +1621,49 @@ point is set."
...
+(define-minor-mode dap-auto-configure-mode
+  "Auto configure dap minor mode."
+  :init-value nil
+  :group 'dap-mode
+  (cond
+   (dap-auto-configure-mode
+    (dap-mode 1)
+    (dap-ui-mode 1)
+    (seq-doseq (feature dap-auto-configure-features)
+      (when-let ((mode (alist-get feature dap-feature->mode)))
```
kiennq, line 1651: With only one binding, you can omit a pair of paren.
```suggestion
      (when-let (mode (alist-get feature dap-feature->mode))
```

dap-mode.el
```diff
@@ -1598,6 +1621,49 @@ point is set."
...
+              (funcall (car mode) 1))
+            (funcall mode 1))))
+    (add-hook 'dap-stopped-hook #'dap--show-feature-window)
+    (add-hook 'dap-terminated-hook #'dap--hide-feature-window))
+   (t
+    (dap-mode -1)
+    (dap-ui-mode -1)
+    (seq-doseq (feature dap-auto-configure-features)
+      (when-let ((mode (alist-get feature dap-feature->mode)))
+        (funcall mode -1)))
```
kiennq, line 1663: I think you need to check the `mode` for `consp` here

ericdallo: nice catch!

dap-mode.el
```diff
@@ -1598,6 +1621,49 @@ point is set."
...
+  (seq-doseq (feature-start-stop dap-auto-configure-features)
+    (when-let ((start-stop (alist-get feature-start-stop dap-feature->start-stop)))
+      (funcall (car start-stop)))))
+
+(defun dap--hide-feature-window (_session)
+  "Hide all debug windows when sessions are dead."
+  (unless (-filter 'dap--session-running (dap--get-sessions))
+    (seq-doseq (feature-start-stop dap-auto-configure-features)
+      (-when-let* ((feature-start-stop (alist-get feature-start-stop dap-feature->start-stop))
+                   (buffer-name (symbol-value (cdr feature-start-stop))))
```
kiennq, line 1638: If this one is meant for buffer name, why don't you assign the buffer name in the `dap-feature->start-stop` as string instead of symbol?

dap-mode.el
```diff
@@ -137,8 +137,31 @@ The hook will be called with the session file and the new set of breakpoint loca
...
+  "Windows to auto show on debugging when in dap-ui-auto-configure-mode."
+  :group 'dap-mode
+  :type '(set (const sessions)
+              (const locals)
+              (const breakpoints)
+              (const expressions)
+              (const controls)
+              (const tooltip)))
+
+(defconst dap-feature->start-stop
```
kiennq, line 153: nit: what's the meaning of `->` here?

ericdallo: It's just that given a feature, it returns the start-stop functions? 
I'm open to suggestions for a better name :sweat_smile:

dap-mode.el
ericdallo: How? The problem is that `dap-feature->start-stop` returns for example `dap-ui--breakpoints-buffer` instead of the value `"*dap-ui-breakpoints*"`, is there any better way to already get the value from the `dap-feature->start-stop`?

dap-mode.el
kiennq: Since this is used to showing windows, how about `dap-features->windows`?

dap-mode.el
kiennq: Ah, I see, I misread `symbol-value` to `symbol-name` 😄

dap-mode.el
ericdallo: Looks good, gonna change :)

yyoncho:
Thank you!

PR #278, opened 2020-05-13
Make dap-ui-locals always expand nodes

ericdallo:
Like VSCode, this PR makes default expand the `dap-ui-locals`

image

image

### DIFF
--- a/dap-ui.el
+++ b/dap-ui.el
@@ -800,7 +800,7 @@ DEBUG-SESSION is the debug session triggering the event."
                                               (dap--cur-session)
                                               variables-reference)))
                   it)
-            (lsp-treemacs-render it " Locals " nil dap-ui--locals-buffer)
+            (lsp-treemacs-render it " Locals " t dap-ui--locals-buffer)
             (or it t))
           (lsp-treemacs-render
            '((:label "Nothing to display..."


yyoncho:
We should expand only one level - expanding all levels will crash emacs when there is cycling ref and will be very slow with a lot of elements.

ericdallo:
I agree, but I didn't find a way to expand only the first level, do you know how?

yyoncho:
It should be implemented in lsp-treemacs.

Check this as well - https://github.com/Alexander-Miller/treemacs/issues/441 - not sure if it will be 100% match for us because we have async threes which makes the task more complex.

ericdallo:
I could add this feature to `lsp-treemacs` but is not in master yet, right?
Should I test it locally and report if works?

yyoncho:
> I could add this feature to `lsp-treemacs` but is not in master yet, right?

What does "not in master" means? IMO it makes sense to change render to accept `nil`/`t`/`number`.

ericdallo:
I mean this treemacs feature is on [another branch](https://github.com/Alexander-Miller/treemacs/tree/depth-expand), not on master branch

yyoncho:
I meant that it should be implemented - it is not present now.

yyoncho:
Sorry I am totally misreading your comments - yes, we should test if treemacs branch is sufficient for us and eventually implement it in lsp-treemacs.

ericdallo:
> Sorry I am totally misreading your comments - yes, we should test if treemacs branch is sufficient for us and eventually implement it in lsp-treemacs.

Perfect :)

PR #282, opened 2020-05-20
dap-feature->modes: remove buffer local mode

kiennq:
`dap-ui-sessions-mode` and `dap-ui-breakpoints-mode` are local mode that should only be enabled in special buffer, right now we enabled it together with `dap-auto-configure-mode`, which cause wrong mode and key bindings enabled in edit buffer. We should fix that.
One example is enabled `dap-auto-configure-mode` in `*scratch*` buffer, and we cannot type `D` anymore, since it's interfere with `dap-ui-breakpoints-mode` key bindings.

Also, fix several byte compile warnings.

### DIFF
--- a/dap-java.el
+++ b/dap-java.el
@@ -226,6 +226,8 @@ initiate `compile' and attach to the process."
   (interactive (list (dap-java--populate-default-args nil)))
   (dap-start-debugging debug-args))
 
+(defvar testng-report-directory)
+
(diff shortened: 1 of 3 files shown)


kiennq:
@yyoncho @ericdallo we should remove `dap-ui-sessions-mode` and `dap-ui-breakpoints-mode` from `dap-feature->modes`, else user will simply not be able to type some keys

PR #285, opened 2020-05-28
dap-features->windows: add REPL window

kiennq:
Allow REPL windows to be shown when enable autoconfigure mode


yyoncho:
Thank you!

PR #286, opened 2020-05-28
Expand first level of dap-ui-locals (#230)

### LINKED ISSUE #230: Feature req: expand first level dap-ui-locals
The new tree widget in `dap-mode` is really nice, but it's lacking in two respects:

- the locals view is always collapsed, forcing me to use the mouse or navigate there to expand it. It would be great (I'd say essential) to have the possibility to expand the first level by default (this is was Eclipse, IDEA do). IMHO, this should be the default behaviour but I'd be happy for a configurable option too :smile: 

- the tree is refreshed without any step being issued. The developer can thus be investigating a deep object graph when `dap-mode` suddenly refreshes the tree, collapsing the whole thing again.

Keep up the good work!

-Torstein

MetaDark:
Fixes #230, depends on https://github.com/emacs-lsp/lsp-treemacs/pull/42

### DIFF
--- a/dap-ui.el
+++ b/dap-ui.el
@@ -806,7 +806,7 @@ DEBUG-SESSION is the debug session triggering the event."
                                               (dap--cur-session)
                                               variables-reference)))
                   it)
-            (lsp-treemacs-render it " Locals " nil dap-ui--locals-buffer)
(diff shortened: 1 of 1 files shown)


ericdallo:
Thank you!

Issue #202, opened 2020-01-27
PYTHONPATH in dap-debug


lccambiaghi:
How can I set the env variable PYTHONPATH when I run `dap-debug`?

My debug template looks like:
```
    (dap-register-debug-template "Debug Python"
                             (list :type "python"
                                   :args "-i"
                                   :cwd (expand-file-name "~/git/projectname")
                                   :env '(("PYTHONPATH" . "/Users/luca/git/projectname/src/"))
                                   :request "launch"
                                   :name "Python :: Run Configuration"))
```

And I get an error in the first lines because of a failed import.

I use `direnv-mode` and in my `.envrc` file I have exported my PYTHONPATH. It seems the env variable is not picked up by `dap-mode`. This works for pytest, the env variables are set like this: https://github.com/wbolster/emacs-python-pytest/blob/master/python-pytest.el#L529-L531 . I tried to replicate the same behaviour with:
```
(after! (direnv dap-python)
  (defvar direnv-non-file-modes)
  (add-to-list 'direnv-non-file-modes 'dap-mode))
```
but it does not work.

Thank you for your help.

yyoncho:
Use this:

``` elisp
(dap-debug
 (list :type "python"
       :args ""
       :cwd nil
       :module nil
       :program "/home/kyoncho/.temp/test.py"
       :request "launch"
       :environment-variables '(("FOO" . "BAR"))
       :name "Python :: Run Configuration"
       :hostName "localhost"
       :host "localhost"))
```

lccambiaghi:
Using your exact template, only replacing `:program` and `:environment-variables`, it does not work. Why did you close the issue?

yyoncho:
Can you put this in `test.py` and then edit `:program` to point to that file and report if it works or not. 

``` python
import os

print(os.environ['FOO'])
```

lccambiaghi:
```
-*- mode: dap-server-log; default-directory: "~/git/emptiesforecast/" -*-
Debug Adapter started at Mon Jan 27 23:42:05

/Users/luca/git/emptiesforecast/.venv/bin/python -m ptvsd --wait --host localhost --port 49947  /Users/luca/git/emptiesforecast/scratch.py 
Traceback (most recent call last):
  File "/Users/luca/.pyenv/versions/3.7.6/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
[...]

PR #293, opened 2020-06-08
Update  dap-python for running in pipenv

MokkeMeguru:
We have many options for executing the python program.
At first, I want to add "pipenv" into the option.

### DIFF
--- a/dap-python.el
+++ b/dap-python.el
@@ -273,7 +273,6 @@ as the pyenv version then also return nil. This works around https://github.com/
 			     (list :type "python-test-at-point"
 				   :args ""
 				   :module "pytest"
-				   :virtualenv "pyenv"
 				   :request "launch"
 				   :virtualenv "pyenv"
 				   :name "Python :: Run pytest (at point)"))
--- a/dap-python.el
+++ b/dap-python.el
@@ -265,7 +265,7 @@ as the pyenv version then also return nil. This works around https://github.com/
                                    :program nil
                                    :module "pytest"
                                    :request "launch"
-				   :virtualenv "pyenv"
+				     :virtualenv "pyenv"
                                    :name "Python :: Run pytest (buffer)"))
 
 (dap-register-debug-provider "python-test-at-point" 'dap-python--populate-test-at-point)
--- a/dap-python.el
+++ b/dap-python.el
@@ -193,9 +193,9 @@ as the pyenv version then also return nil. This works around https://github.com/
          (debug-port (dap--find-available-port))
 	 (virtualenv-type (plist-get conf :virtualenv))
          (python-executable 
-	  (cond ((= virtualenv-type "pyenv")
+	  (cond ((string= virtualenv-type "pyenv")
 		 (dap-python--pyenv-executable-find dap-python-executable))
-		((= virtualenv-type "pipenv")
+		((string= virtualenv-type "pipenv")
 		 (dap-python--pipenv-executable-find dap-python-executable))
 		(t
 		 (executable-find dap-python-executable))))
--- a/dap-python.el
+++ b/dap-python.el
@@ -255,7 +255,7 @@ as the pyenv version then also return nil. This works around https://github.com/
                                    :module nil
                                    :program nil
                                    :request "launch"
-				   :virtualenv "pyenv"
+				     :virtualenv "pyenv"
                                    :name "Python :: Run file (buffer)"))
 
 (dap-register-debug-template "Python :: Run pytest (buffer)"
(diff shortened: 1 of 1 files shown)


dap-python.el
```diff
@@ -246,14 +265,17 @@ as the pyenv version then also return nil. This works around https://github.com/
...
                                    :name "Python :: Run pytest (buffer)"))
 
 (dap-register-debug-provider "python-test-at-point" 'dap-python--populate-test-at-point)
 (dap-register-debug-template "Python :: Run pytest (at point)"
 			     (list :type "python-test-at-point"
 				   :args ""
 				   :module "pytest"
+				   :virtualenv "pyenv"
 				   :request "launch"
+				   :virtualenv "pyenv"
```
ericdallo, line 278: Duplicated?

dap-python.el
```diff
@@ -237,6 +255,7 @@ as the pyenv version then also return nil. This works around https://github.com/
                                    :module nil
                                    :program nil
                                    :request "launch"
+				   :virtualenv "pyenv"
```
ericdallo, line 258:
```suggestion
				     :virtualenv "pyenv"
```

dap-python.el
```diff
@@ -246,14 +265,17 @@ as the pyenv version then also return nil. This works around https://github.com/
                                    :program nil
                                    :module "pytest"
                                    :request "launch"
+				   :virtualenv "pyenv"
```
ericdallo, line 268:
```suggestion
				     :virtualenv "pyenv"
```

dap-python.el
```diff
@@ -180,7 +191,14 @@ as the pyenv version then also return nil. This works around https://github.com/
   "Populate CONF with the required arguments."
   (let* ((host "localhost")
          (debug-port (dap--find-available-port))
-         (python-executable (dap-python--pyenv-executable-find dap-python-executable))
+	 (virtualenv-type (plist-get conf :virtualenv))
+         (python-executable 
+	  (cond ((= virtualenv-type "pyenv")
```
ericdallo, line 196: I think `string=` is better for string comparassion

MokkeMeguru: I afraid that symbol will pollute global variables... (I know emacs-lisp a bit.)
string= is better

yyoncho:
@mpanarin, willing to take a look?

dap-python.el
MokkeMeguru: Thanks!

dap-python.el
MokkeMeguru: If the symbol is correct in terms of emacs lisp, please tell me.

mpanarin:
It is the same as https://github.com/emacs-lsp/dap-mode/issues/202
To be honest, I am against explicit support of the virtualenv in `lsp-mode` or `dap-mode`.
This should be handled by emacs and its `executable-find`. There are a bunch of packages and minor-modes that will make `executable-find` find a proper python in virtualenv. I mentioned a few in the issue above.

The problem with supporting virtualenvs is that everyone uses them differently. You can use them through `pipenv`, 'virtualenv', 'virtualenvwrapper', 'poetry' to name a few. And I don't think we should write support for all of them.
`pyenv` is a bit different as it is not really a virtualenv, so we have a diret support for it.

ericdallo:
I agree with @mpanarin, IMO we should make dap-mode extensible and easy to setup with different templates, support all possible virtualenvs seems to be too much.

MokkeMeguru:
But now you are supported pyenv. If you want to do so, you should remove pyenv option, shouldn't you?

MokkeMeguru:
In my opinion, we can give the execute-path into a template.
Can you fix it?

PanarinM:
@MokkeMeguru pyenv is a version manager, not a virtualenv manager.
As I said, it is a different story, so we support it explicitly

mpanarin:
@MokkeMeguru pyenv is a version manager, not a virtualenv manager.
As I said, it is a different story, so we support it explicitly

MokkeMeguru:
If so, then we must consider conda, and Pyflow... It will be too complex problem...

greboide:
So how to run dap on python with pipenv? No way?

dap-python.el
nbfalcon: Symbols don't pollute the global namespace, unless you intern them, which you needn't do. A symbol ('foo) is just a lisp object like an int, string, .... You could use (eq virtualenv-type 'pyenv) instead of (= virtualenv-type "pyenv"), the latter invalid anyway since = can only compare ints and markers.

htgoebel:
@MokkeMeguru @mpanarin Since this pull-request is not accepted and #294 is an alternative: can this one be closed?

MokkeMeguru:
hi @htgoebel 
IMO, this pr can be closed.
thanks for your review

PR #299, opened 2020-06-15
Update link to lldb-vscode

timfel:
The old repository is obsolete and the code doesn't compile anymore with llvm-10


ericdallo:
Thank you!

ericdallo:
@timfel actually, you need to fix it [here](https://github.com/emacs-lsp/dap-mode/blob/master/docs/page/configuration.md#lldb) otherwise, the changes will be overwritten in a next commit.

PR #300, opened 2020-06-15
Update link to lldb-vscode

timfel:
The old repository is obsolete and the code doesn't compile anymore with llvm-10.


timfel:
Thanks @ericdallo for the pointer

PR #301, opened 2020-06-16
Use lsp-protocol

ericdallo:
https://github.com/emacs-lsp/lsp-mode/issues/1751

### DIFF
--- a/dap-java.el
+++ b/dap-java.el
@@ -93,10 +93,10 @@ If the port is taken, DAP will try the next port."
   "Get class FDQN."
   (-if-let* ((symbols (lsp--get-document-symbols))
              (package-name (-some->> symbols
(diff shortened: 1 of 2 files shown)


dap-java.el
```diff
@@ -93,30 +93,30 @@ If the port is taken, DAP will try the next port."
   "Get class FDQN."
   (-if-let* ((symbols (lsp--get-document-symbols))
              (package-name (-some->> symbols
-                             (-first (-lambda ((&hash "kind")) (= kind 4)))
-                             (gethash "name")))
+                             (-first (-lambda ((&DocumentSymbol :kind)) (= kind 4)))
```
yyoncho, line 96: it is a good time to replace those with the corresponding constants.

ericdallo: I forgot we already have constants for those, I'll replace them

ericdallo: Done!

PR #307, opened 2020-06-25
fix regression for window-point-insertion-type=t being overwritten

skebanga:
output buffer now follows when new output is added


yyoncho:
Thank you for contributing to `dap-mode` and congrats for your first PR in `emacs-lsp` organization! :tada:

Issue #324, opened 2020-08-06
Integrate my launch.json plugin


nbfalcon:
launch.json is the de-facto standard for debugger launch configurations, and as such dap-mode should support that.
I have written a plugin (https://github.com/nbfalcon/dap-mode-launch-json) to add that feature, and would love to have it merged!

Feedback is welcome, and I would be happy to provide assistance if needed.

yyoncho:
This is something definitely on the roadmap of dap-mode. I will try to provide a feedback for the plugin and how do I see the integration with dap-mode. You may turn it into a PR to make leaving the feedback easier to track.

ericdallo:
Nice!
I'd like to see a PR adding this feature too :)

nbfalcon:
My current concerns are the following:
- I define functions named projectile-*, which I should probably try to get into mainline projectile. Here I have a question: does dap-mode depend on some specific minimum version of projectile? If so, then how should we handle this?
- I define various functions named buffer-*; these are buffer filename related utilities, and I am unsure where I should try to merge them into

ericdallo:
@nbfalcon it's not good to have functions in you lib that doesn't has you lib's name prefix, to avoid other libs with same function name, so you may rename it adding your lib prefix name.
If this comes to `dap-mode`, we may have the prefix `dap-buffer-*` and etc

nbfalcon:
Ok, I will do that. But I want to get the projectile-* into mainline projectile, and I am currently writing the PR for that. However, that would mean that, if my plugin were merged, dap-mode would have to depend on a very new version of projectile. Would there be any problems with that?

ericdallo:
I don't think so, it's pretty common we merge fixes and features on other projects and lsp-mode use it, it's not the best approach, but we recommend the user to update the package to the latest if face any problem.
What we can do to is check if the function is defined(if not, then the projectile is outdated) and not offer the feature, despite that this looks a major feature for dap-mode, so I'm not sure we should do that

yyoncho:
dap-mode should not depend on projectile. There is lsp-workspace-root function which you should use for finding the project root.

yyoncho:
The rest of `projectile-*` in your package functions are present in `f` package.

nbfalcon:
Ok. I will update my plugin to not use projectile.

nbfalcon:
Can we depend on f?

yyoncho:
We already depend on `f`.

nbfalcon:
[...]

PR #325, opened 2020-08-06
launch.json support for dap-mode

nbfalcon:
PR for #324.
The change to dap-ui.el was already merged, and I have no idea how to remove it. I am still a bit new to git.

### DIFF
--- a/dap-launch.el
+++ b/dap-launch.el
@@ -1,0 +1,241 @@
+;; dap-mode-launch-json.el --- support launch.json -*- lexical-binding: t -*-
+
+;; Copyright (C) 2020 Nikita Bloshchanevich
+
+;; Author: Nikita Bloshchanevich <nikblos@outlook.com>
+;; Keywords: languages
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+(require 'dap-mode)
+(require 'lsp)
+(require 'cl-lib)
+(require 'f)
+
+;;; Commentary:
+;; Extend dap-mode with support for launch.json files
+
+;;; Code:
+
+(defun dap--project-find-launch-json ()
+  "Return the location of the launch.json file in the current project."
+  (when-let ((project (lsp-workspace-root)))
+    (concat project "/launch.json")))
+
+(defun dap--project-get-launch-json ()
+  "Parse the project's launch.json as json data and return the result."
+  (when-let ((launch-json (dap--project-find-launch-json))
+             (json-object-type 'plist))
+    (json-read-file launch-json)))
+
+(defun dap--parse-launch-json (json)
+  "Return a list of all launch configurations in JSON.
+JSON must have been acquired with `dap--project-get-launch-json'."
+  (or (plist-get json :configurations) (list json)))
+
+(defun dap--project-parse-launch-json ()
+  "Return a list of all launch configurations for the current project."
+  (dap--parse-launch-json (dap--project-get-launch-json)))
+
+(defun dap--configuration-get-name (conf)
+"Return the name of launch configuration CONF."
+  (plist-get conf :name))
+
+(defun dap--project-project-basename (&optional dir)
+  "Return the name of the project root directory.
+Starts the project-root search at DIR."
+    (file-name-nondirectory (directory-file-name (lsp-workspace-root dir))))
+
+(defun dap--project-relative-file (&optional file dir)
+  "Return the path to FILE relative to the project root.
+The search for the project root starts at DIR. FILE defaults to
+variable `buffer-file-name'."
+    (f-relative (or file buffer-file-name) (lsp-workspace-root dir)))
+
+(defun dap--project-relative-dirname (&optional file dir)
+  "Return the path to the directory of file relative to the project root.
+The search for the project root starts at DIR. FILE defaults to
+variable `buffer-file-name'"
+  (dap--project-relative-file (file-name-directory (or file buffer-file-name))
+                              dir))
+
+(defun dap--buffer-basename ()
+  "Return the name of the current buffer's file without its directory."
+  (file-name-nondirectory buffer-file-name))
+
+(defun dap--buffer-basename-sans-extension ()
+  "Same as `dap--buffer-basename', but without the extension."
+  (file-name-sans-extension (dap--buffer-basename)))
+
+(defun dap--buffer-extension ()
+  "Return the extension of the buffer's file with a leading dot.
+If there is either no file associated with the current buffer or
+if that file has no extension, return the empty string."
+  (if-let ((buffer-name buffer-file-name)
+           (ext (file-name-extension buffer-name)))
+      (concat "." ext)
+    ""))
+
+(defun dap--buffer-dirname ()
+  "Return the directory the buffer's file is in."
+  (file-name-directory buffer-file-name))
+
+(defun dap--buffer-current-line ()
+  "Return the line the cursor is on in the current buffer."
+  (number-to-string (line-number-at-pos)))
+
+(defun dap--buffer-selected-text ()
+  "Return the text selected in the current buffer.
+If no text is selected, return the empty string."
+  ;; Cannot fail, as if there is no mark, (mark) and (point) will be equal, and
+  ;; (`buffer-substring-no-properties') will yield "", as it should.
+  (buffer-substring-no-properties (mark) (point)))
+
+(defun dap--launch-json-warn-nil (text)
+  "Issue a warning related to the launch.json file containing TEXT.
+Always return nil."
+  (message (concat "warning: launch.json: " text))
+  nil)
+
+(defun dap--warn-unknown-envvar (var)
+  "Warn, related to launch.json, that VAR is an unknown environment variable.
+Always return nil."
+  (dap--launch-json-warn-nil (format "no such environment variable '%s'" var)))
+
+(defun dap--launch-json-getenv (var)
+  "Return the environment variable in matched string VAR.
+Only for use in `dap-launch-json-variables'."
+  (let ((envvar (match-string 1 var)))
+    (or (getenv envvar) (dap--warn-unknown-envvar envvar) "")))
+
+(defvar dap-launch-json-variables
+  ;; list taken from https://code.visualstudio.com/docs/editor/variables-reference
+  '(("workspaceFolderBasename" . dap--project-project-basename)
+    ("workspaceFolder" . lsp-workspace-root)
+    ("relativeFileDirname" . dap--project-relative-dirame)
+    ("relativeFile" . dap--project-relative-file)
+    ("fileBasenameNoExtension" . dap--buffer-basename-sans-extension)
+    ("fileBasename" . dap--buffer-basename)
+    ("fileDirname" . dap--buffer-dirname)
+    ("fileExtname" . dap--buffer-extension)
+    ("lineNumber" . dap--buffer-current-line)
+    ("selectedText" . dap--buffer-selected-text)
+    ("file" . buffer-file-name)
+    ("env:\\(.*\\)" . dap--launch-json-getenv)
+    ;; technically not in VSCode, but I still wanted to add a way to escape $
+    ("$" . "$")
+    ;; the following variables are valid in VSCode, but have no meaning in
+    ;; Emacs, and are as such unsupported.
+    ;; ("cwd") ;; the task runner's current working directory,
+    ;;         ;; not `default-directory'
+    ;; ("execPath")
+    ;; ("defaultBuildTask")
+    )
+  "Alist of (REGEX . VALUE) pairs listing variables usable in launch.json files.
+This list is iterated from the top to bottom when expanding
+variables in the strings of the selected launch configuration
+from launch.json or in `dap-expand-variable'.
+
+When a REGEX matches (`string-match'), its corresponding VALUE is
+evaluated as follows: if it is a function (or a quoted lambda),
+that function is called with `funcall', and its result, which
+must be a string, is used in place of the variable. If you used
+capture groups in REGEX, the function you specified in VALUE is
+called with the variable as its only argument. This way, you can
+use `string-match' to get the capture groups. If, however, REGEX
+does not contain capture groups, your function is called without
+any arguments. Otherwise, if it is a symbol, the symbol's value
+is used the same way. Lastly, if it is a string, the string is
+used as a replacement. If no regex matches, the empty string is
+used as a replacement and a warning is issued.
+
+See `dap--launch-json-getenv' for an example on how to use
+capture groups in REGEX.")
+
+(defun dap--launch-json-eval-poly-type (value var)
+  "Get the value from VALUE depending on its type.
+If it is a function, and VAR is not nil, call VALUE and pass VAR as an argument.
+If it is a symbol, return its value.
+Otherwise, return VALUE"
+  (cond ((and var (functionp value)) (funcall value var))
+        ((functionp value) (funcall value))
+        ((symbolp value) (symbol-value value))
+        (t value)))
+
+(defun dap--warn-var-nil (var)
+  "Warn that VAR is an unknown launch.json variable."
+  (dap--launch-json-warn-nil
+   (format "variable '%s' is unknown and was ignored" var)))
+
+(defun dap-expand-variable (var)
+  "Expand variable VAR using `dap--launch-json-variables'."
+  (catch 'ret
+    (save-match-data
+      (dolist (var-pair dap-launch-json-variables)
+        (when (string-match (car var-pair) var)
+          (throw 'ret
+                 (or
+                  (dap--launch-json-eval-poly-type
+                   (cdr var-pair)
+                   (if (= (length (match-data)) 2) ;; no capture groups
+                       nil
+                     var
+                     (message "wow capture groups")))
+                  (dap--warn-var-nil var)
+                  "")))))
+    nil))
+
+(defun dap-expand-variables-in-string (s)
+  "Expand all launch.json variables of the from ${variable} in S.
+Return the result."
+  (let ((old-buffer (current-buffer)))
+    (with-temp-buffer
+      (insert s)
+      (goto-char (point-min))
+
+      (save-match-data
+        (while (re-search-forward "${\\([^}]*\\)}" nil t)
+          (let ((var (match-string 1)))
+            (replace-match
+             (with-current-buffer old-buffer
+               (dap-expand-variable var))))))
+
+      (buffer-string))))
+
+(defun dap--launch-json-expand-vars (conf)
+  "Non-destructively expand all variables in all strings of CONF.
+CONF is regular dap-mode launch configuration. Return the result."
+  (cond ((listp conf)
+         (apply #'nconc (cl-loop for (k v) on conf by #'cddr collect
+                                 (list k (dap--launch-json-expand-vars v)))))
+        ((stringp conf) (dap-expand-variables-in-string conf))
+        (t conf)))
+
+(defun dap--launch-json-prompt-configuration ()
+  "Prompt the user to select a launch configuration from the launch.json file."
+  (dap--completing-read "Select configuration: "
+                        (dap--project-parse-launch-json)
+                        #'dap--configuration-get-name))
+
+(defun dap--acquire-launch-json ()
+  "Prompt the user for a launch configuration and expand its variables."
+  (dap--launch-json-expand-vars (dap--launch-json-prompt-configuration)))
+
+(defun dap-debug-launch-json ()
+  "Read the project's launch.json and ask the user for a launch configuration."
+  (interactive)
+  (dap-debug (dap--acquire-launch-json)))
+
+(provide 'dap-mode-launch-json)
+;;; dap-mode-launch-json.el ends here
--- a/dap-launch.el
+++ b/dap-launch.el
@@ -102,7 +102,7 @@ If no text is selected, return the empty string."
   ;; (`buffer-substring-no-properties') will yield "", as it should.
   (buffer-substring-no-properties (mark) (point)))
 
-(defun dap--launch-json-getenv (var)
+(defun dap--launch-configuration-var-getenv (var)
   "Return the environment variable in matched string VAR.
 Only for use in `dap-launch-configuration-variables'."
   (let ((envvar (match-string 1 var)))
@@ -124,7 +124,7 @@ Only for use in `dap-launch-configuration-variables'."
     ("lineNumber" . dap--buffer-current-line)
     ("selectedText" . dap--buffer-selected-text)
     ("file" . buffer-file-name)
-    ("env:\\(.*\\)" . dap--launch-json-getenv)
+    ("env:\\(.*\\)" . dap--launch-configuration-var-getenv)
     ;; technically not in VSCode, but I still wanted to add a way to escape $
     ("$" . "$")
     ;; the following variables are valid in VSCode, but have no meaning in
@@ -135,8 +135,7 @@ Only for use in `dap-launch-configuration-variables'."
     ;; ("defaultBuildTask")
     )
   "Alist of (REGEX . VALUE) pairs listing variables usable in launch.json files.
-This list is iterated from the top to bottom when expanding
-variables in the strings of the selected launch configuration
+This list is iterated from the top to bottom when expanding variables in the strings of the selected launch configuration
 from launch.json or in `dap-expand-launch-configuration-variable'.
 
 When a REGEX matches (`string-match'), its corresponding VALUE is
@@ -152,10 +151,10 @@ is used the same way. Lastly, if it is a string, the string is
 used as a replacement. If no regex matches, the empty string is
 used as a replacement and a warning is issued.
 
-See `dap--launch-json-getenv' for an example on how to use
+See `dap--launch-configuration-var-getenv' for an example on how to use
 capture groups in REGEX.")
 
-(defun dap--launch-json-eval-poly-type (value var)
+(defun dap--eval-poly-type (value var)
   "Get the value from VALUE depending on its type.
 If it is a function, and VAR is not nil, call VALUE and pass VAR as an argument.
 If it is a symbol, return its value.
@@ -173,7 +172,7 @@ Otherwise, return VALUE"
         (when (string-match (car var-pair) var)
           (throw 'ret
                  (or
-                  (dap--launch-json-eval-poly-type
+                  (dap--eval-poly-type
                    (cdr var-pair)
                    (if (= (length (match-data)) 2) ;; no capture groups
                        nil
--- a/dap-launch.el
+++ b/dap-launch.el
@@ -104,14 +104,14 @@ If no text is selected, return the empty string."
 
 (defun dap--launch-json-getenv (var)
   "Return the environment variable in matched string VAR.
-Only for use in `dap-launch-json-variables'."
+Only for use in `dap-launch-configuration-variables'."
   (let ((envvar (match-string 1 var)))
     (or (getenv envvar)
         (lsp-warn "launch.json: no such environment variable '%s' (in ${%s})"
                   envvar var)
          "")))
 
-(defvar dap-launch-json-variables
+(defvar dap-launch-configuration-variables
   ;; list taken from https://code.visualstudio.com/docs/editor/variables-reference
   '(("workspaceFolderBasename" . dap--project-project-basename)
     ("workspaceFolder" . lsp-workspace-root)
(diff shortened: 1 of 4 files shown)


dap-launch.el
```diff
@@ -0,0 +1,241 @@
+;; dap-mode-launch-json.el --- support launch.json -*- lexical-binding: t -*-
+
+;; Copyright (C) 2020 Nikita Bloshchanevich
+
+;; Author: Nikita Bloshchanevich <nikblos@outlook.com>
+;; Keywords: languages
```
ericdallo, line 6: as this file is part of the `dap-mode` package, I think ẃe don't need these 2 lines
```suggestion
```

dap-launch.el
```diff
@@ -0,0 +1,241 @@
...
+  "Prompt the user for a launch configuration and expand its variables."
+  (dap--launch-json-expand-vars (dap--launch-json-prompt-configuration)))
+
+(defun dap-debug-launch-json ()
+  "Read the project's launch.json and ask the user for a launch configuration."
+  (interactive)
+  (dap-debug (dap--acquire-launch-json)))
+
+(provide 'dap-mode-launch-json)
+;;; dap-mode-launch-json.el ends here
```
ericdallo, line 241:
```suggestion
(provide 'dap-launch-json)
;;; dap-launch-json.el ends here
```

dap-launch.el
```diff
@@ -0,0 +1,241 @@
+;; dap-mode-launch-json.el --- support launch.json -*- lexical-binding: t -*-
```
ericdallo, line 1:
```suggestion
;; dap-launch-json.el --- support launch.json -*- lexical-binding: t -*-
```

dap-launch.el
```diff
@@ -0,0 +1,241 @@
...
+      (save-match-data
+        (while (re-search-forward "${\\([^}]*\\)}" nil t)
+          (let ((var (match-string 1)))
+            (replace-match
+             (with-current-buffer old-buffer
+               (dap-expand-variable var))))))
+
+      (buffer-string))))
+
+(defun dap--launch-json-expand-vars (conf)
```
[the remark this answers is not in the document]
nbfalcon: Actually, we can! I will rename the function to dap-launch-configuration-expand-vars (no longer private, perhaps users may want this?) and we can just call (setq launch-args (dap-launch-configuration-expand-vars launch-configuration)) at the start of dap-start-debugging. That would mean that regular templates could use variables (e.g.
```elisp
(dap-register-template ... :executable "${workspaceFolder}/build/gtk-print")
```).

However, merging it into dap-mode.el might not be a good idea. I think having it as a separate file (e.g. dap-variables.el) would be better, because all the variable-code would be in one place and it would make it easier to add new variables. What do you think? Also, how could we load it from dap-mode.el?

File: dap-launch.el [hunk 5]
```diff
@@ -0,0 +1,241 @@
...
+    ("fileExtname" . dap--buffer-extension)
+    ("lineNumber" . dap--buffer-current-line)
+    ("selectedText" . dap--buffer-selected-text)
+    ("file" . buffer-file-name)
+    ("env:\\(.*\\)" . dap--launch-json-getenv)
+    ;; technically not in VSCode, but I still wanted to add a way to escape $
+    ("$" . "$")
+    ;; the following variables are valid in VSCode, but have no meaning in
+    ;; Emacs, and are as such unsupported.
+    ;; ("cwd") ;; the task runner's current working directory,
```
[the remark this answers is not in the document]
nbfalcon: This is a bit of a misnomer by microsoft. "cwd" has nothing to do with the program's current working directory, but is in fact "the task runner's current working directory" (see here: https://code.visualstudio.com/docs/editor/variables-reference, also here https://code.visualstudio.com/docs/editor/tasks for what a task runner is). Since there is no dap-mode task runner, this variable does not make sense.

dap-launch.el
```diff
@@ -0,0 +1,241 @@
...
+(defun dap--launch-json-expand-vars (conf)
+  "Non-destructively expand all variables in all strings of CONF.
+CONF is regular dap-mode launch configuration. Return the result."
+  (cond ((listp conf)
+         (apply #'nconc (cl-loop for (k v) on conf by #'cddr collect
+                                 (list k (dap--launch-json-expand-vars v)))))
+        ((stringp conf) (dap-expand-variables-in-string conf))
+        (t conf)))
+
+(defun dap--launch-json-prompt-configuration ()
```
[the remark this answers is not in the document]
nbfalcon: I agree. My approach is a bit ad-hoc. I propose that we add a plugin architecture to dap-debug. It should then call something like (dap-gather-templates) instead of using dap-debug-template-configurations in the interactive block. That function would then iterate over something like dap-backends-alist and gather all templates. I don't see any other references to that variable (aside from defcustoms) so that shouldn't be a problem?

dap-launch.el
```diff
@@ -0,0 +1,241 @@
...
+      (dolist (var-pair dap-launch-json-variables)
+        (when (string-match (car var-pair) var)
+          (throw 'ret
+                 (or
+                  (dap--launch-json-eval-poly-type
+                   (cdr var-pair)
+                   (if (= (length (match-data)) 2) ;; no capture groups
+                       nil
+                     var
+                     (message "wow capture groups")))
```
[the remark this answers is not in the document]
nbfalcon: That is rather embarrassing.

dap-launch.el
```diff
@@ -0,0 +1,241 @@
...
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+(require 'dap-mode)
+(require 'lsp)
```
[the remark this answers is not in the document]
nbfalcon: Ok, fixed it!

dap-launch.el
```diff
@@ -0,0 +1,241 @@
...
+  (number-to-string (line-number-at-pos)))
+
+(defun dap--buffer-selected-text ()
+  "Return the text selected in the current buffer.
+If no text is selected, return the empty string."
+  ;; Cannot fail, as if there is no mark, (mark) and (point) will be equal, and
+  ;; (`buffer-substring-no-properties') will yield "", as it should.
+  (buffer-substring-no-properties (mark) (point)))
+
+(defun dap--launch-json-warn-nil (text)
```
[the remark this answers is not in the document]
nbfalcon: Fixed!

dap-launch.el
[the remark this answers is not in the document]
yyoncho: > However, merging it into dap-mode.el might not be a good idea. I think having it as a separate file (e.g. dap-variables.el) would be better, because all the variable-code would be in one place and it would make it easier to add new variables. What do you think? Also, how could we load it from dap-mode.el?

If you want to structure that as a utility - then I guess it is fine to reverse the dependency. There is also dap-util.el which hosts vscode related stuff atm.

dap-launch.el
nbfalcon: Fixed in current master.

dap-launch.el
nbfalcon: Fixed!

nbfalcon:
How do you load files from modules?

dap-launch.el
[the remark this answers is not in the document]
nbfalcon: That was not the idea, as variable expansion should be integrated into normal configurations. Rather, dap-launch.el would be loaded from dap-mode.el. Or should it better be in dap-mode.el?

dap-launch.el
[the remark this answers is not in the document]
yyoncho: Thank you for clearing that up. As a side note, I am planning to create tasks extensions for emacs.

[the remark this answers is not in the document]
nbfalcon: I will add it as a variable if its done then.

dap-launch.el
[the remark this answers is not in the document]
yyoncho: I am usually against introducing additional layers of abstraction when it is not clear what the benefit is but I guess in this case having something like won't affect the complexity of the code. 
```
(-mapcat #'funcall dap-templates-functions)
```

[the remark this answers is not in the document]
nbfalcon: Yes, that was my idea. I will implement this.

yyoncho:
can you elaborate? What I am suggesting is having the code either in
dap-mode.el (which IMO is fine) or having the code in dap-FOO.el and
require that from dap-mode.el. In order that to happen dap-FOO.el should
not require dap-mode.el.

nbfalcon:
dap-launch.el will be split up (I already did that as part of a separate branch) into two files: dap-launch.el, which will contain the code to provide a json backend and into dap-variables.el, which will contain the code for expanding variables. Both of them won't have to depend on dap-mode.el

dap-launch.el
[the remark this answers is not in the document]
nbfalcon: Implemented in latest master of my fork.

nbfalcon:
Please check out my latest commit to master: I have split up dap-launch.el into dap-variables.el. dap-launch.el now only contains code for parsing launch.json files, and dap-variables.el handles variable expansion. dap-launch.el has a function that can be used as a dap launch configuration provider for the architecture implemented in e711698. Both files do not depened on dap.

dap-mode.el
```diff
@@ -1442,13 +1442,22 @@ If ORIGIN is t, return the original configuration without prepopulation"
           (error "There is no debug provider for language %s"
                  (or (plist-get debug-args :type) "'Not specified'"))))))
 
+(defun dap-debug-template-configurations-provider ()
+  dap-debug-template-configurations)
```
yyoncho, line 1446: launch.json params should be on by default.

dap-launch.el
```diff
@@ -0,0 +1,56 @@
+;; dap-mode.el --- support launch.json -*- lexical-binding: t -*-
```
yyoncho, line 1: dap-launch.

Also, all variables/functions should be prefixed with dap-launch/dap-variables

dap-mode.el
nbfalcon: Don't we need to load dap-launch.el first?

dap-launch.el
nbfalcon: Fixed both issues in latest master.

dap-mode.el
yyoncho: yes we need unless the function is marked as autoloaded.

nbfalcon: How do you do that?

nbfalcon: Fixed in latest master.

nbfalcon:
I tried using this patched dap-mode on one of my projects and it works. The launch configurations extracted from its launch.json file are shown in dap-debug. If there is no launch.json, that is handled gracefully without any warnings orerrors and no launch configurations are added in that case. This can now probably be merged and I will write documentation on how to use it.

dap-launch.el
```diff
@@ -0,0 +1,64 @@
...
+Extract the name from the :name property."
+  (push (dap-launch-configuration-get-name conf) conf))
+
+(defun dap-launch-parse-launch-json (json)
+  "Return a list of all launch configurations in JSON.
+JSON must have been acquired with `dap-launch--get-launch-json'."
+  (mapcar #'dap-launch-configuration-prepend-name
+          (or (plist-get json :configurations) (list json))))
+
+(defun dap-launch-find-parse-launch-json ()
```
kiennq, line 57: This function needed to be marked as autoload
```suggestion
;;;###autoload
(defun dap-launch-find-parse-launch-json ()
```

dap-variables.el
```diff
@@ -0,0 +1,187 @@
...
+does not contain capture groups, your function is called without
+any arguments. Otherwise, if it is a symbol, the symbol's value
+is used the same way. Lastly, if it is a string, the string is
+used as a replacement. If no regex matches, the empty string is
+used as a replacement and a warning is issued.
+
+See `dap--launch-configuration-var-getenv' for an example on how to use
+capture groups in REGEX.")
+
+(defun dap--variables-eval-poly-type (value var)
```
kiennq, line 131:
```suggestion
(defun dap-variables--eval-poly-type (value var)
```

nbfalcon: Fixed in latest master.

dap-variables.el
```diff
@@ -0,0 +1,187 @@
+;;; dap-variables.el --- Launch configuration variables -*- lexical-binding: t -*-
```
kiennq, line 1: Does it make sense to move this file into `dap-utils`?
This file is mostly utilities too.

nbfalcon: Perhaps the functions could be included in dap-utils? But the purpose of the file is to define variables that can be expanded in launch configurations, and all the functions serve this purpose. That is, they implement variables that can be expanded.
By variables I mean that if you have a launch configuration like this (taken from my gtk-print project)
```json
{
    "configurations": [
        {
            "name": "Debug",
            "type": "gdb",
            "request": "launch",
            "target": "${workspaceFolder}/build/gtk-print",
            "debugger_args": [ "-ex", "set args /tmp/printme.pdf" ]
        },
        {
            "name": "Debug (attach to localhost:4596)",
            "type": "gdb",
            "request": "attach",
            "executable": "${workspaceFolder}/build/gtk-print",
            "target": ":4596",
            "remote": true
        }
    ]
}
```
Things like ${workspaceFolder} would be expanded by calling the appropriate functions in dap-variable-launch-configuration-variables. This works in normal launch configurations in lisp form too.

kiennq:
Although no need to address in this PR, I think these kind of configurations reading from file and be benefitted from `lsp-protocol`.
Since the schema is fixed, the protocol for those configurations can be defined, when we use that it will be much cleaner with type supports.

nbfalcon:
Can you elaborate further on how I could integrate this? And is it worth it? Because launch.json files are actually very simple:
Either the root object is the only configuration, or, if it has a "configurations" key, the value of that key is a list of launch configurations. What I could see being an issue here is that it is either "configurations" or the root object. That is, a launch.json file like this would be entirely valid and as such must be supported, too:
```json
{
    "name": "Debug",
    "executable": "${workspaceFolder}/build/gtk-print",
    "type": "gdb"
}
```

nbfalcon:
Also, the schema isn't entirely fixed: each individual launch configuration can have whatever keys it wants to have and these may or may not be supported by some debug adapter. That is, there is no set of valid keys. All dap-launch.el does is parse the launch.json file and pass the launch configurations (plists due to json-object-type) on directly do dap-debug. Strings are then expanded in dap-start-debugging.

kiennq:
> Also, the schema isn't entirely fixed: each individual launch configuration can have whatever keys it wants to have and these may or may not be supported by some debug adapter. That is, there is no set of valid keys. All dap-launch.el does is parse the launch.json file and pass the launch configurations (plists due to json-object-type) on directly do dap-debug. Strings are then expanded in dap-start-debugging.

Ok, if the schema for each one is not fixed, then no need to force use `lsp-protocol` then.

> Can you elaborate further on how I could integrate this? And is it worth it? Because launch.json files are actually very simple:
> Either the root object is the only configuration, or, if it has a "configurations" key, the value of that key is a list of launch configurations. What I could see being an issue here is that it is either "configurations" or the root object.

If you want to use `lsp-protocol` this issue can be as simple as a check of `lsp-dap-xyz?` on type in a `cond` expression.
More changes will be needed so that `plist-get` will be replace with appropriate terms too. 


@yyoncho I heard that there's plan to move `dap-mode` onto `lsp-mode` improved json parser?

yyoncho:
> @yyoncho I heard that there's plan to move `dap-mode` onto `lsp-mode` improved json parser?

If you are asking for lsp-protocol.el - I have no plans to do that switch. If you are asking about the parsing code - we could do that at some point, but it is not something urgent because dap-mode has no perf issues.

kiennq:
I also approve this, can you check why CI seems not running in this?

ericdallo:
For some reason, travis is not running anymore for `dap-mode`... we should migrate it to use GithubActions like `lsp-mode`

nbfalcon:
Awesome! Thank you all!

yyoncho:
Few thoughts:

1. I think that it will be great if we can navigate to the template location in the json file. That could happen from a simple UI like the one vscode has. 
2. It will be great if we can support auto-completion in that file. ATM we have json language server support and it support schemas - I am not sure how powerfull json schemas are and whether we will be able to generate a schema for the current template. 
3. Support for templates defined in package.json.

nbfalcon:
1. That looks like a cool feature. Is there some way to parse JSON with cursor positions (each object also has to stores its position in the file)? Perhaps with an external tool like jq?
2. I will look into JSON schemas.
3. Could that really work? I think you're referring to "scripts" : {...}. But they can be arbitrarily complex, involve pipes, etc.. How could we map that to launch configurations? It is unclear what is to be debugged, with what adapter and with what parameters. From skimming one of VSCode's pages (https://code.visualstudio.com/docs/nodejs/nodejs-debugging) it appears that even Microsoft gave up on that feature, and recommends to just manually translate "scripts" sections to launch.json

nbfalcon:
One more question: how could I get the latest dap-mode (current master) in spacemacs?

yyoncho:
> One more question: how could I get the latest dap-mode (current master) in spacemacs?

`SPC f e U`

FYI, to make spacemacs work with the locally checkouted files I do that:

 `(shell-command "ln -s ~/Sources/lsp/dap-mode/*el /home/yyoncho/.emacs.d/elpa/27.1/develop/dap-mode-*/ -fv")`

> I think you're referring to "scripts" : {...}.

No, this - https://github.com/golang/vscode-go/blob/master/package.json#L375

nbfalcon:
In that case we would need another dap-mode backend, e.g. in dap-configuration-snippets.el, and we would need to add more variables. I will implement that. Is there a way to get Emacs' JSON parser to include line number information? Because the thought of writing many jq (for the line numbers, goto launch configuration, feature 1.) scripts and elisp adapters to support lots of different launch configuration formats does not excite me.

nbfalcon:
Is there a function to do mapcar, but with a body instead of a function? I.e. is there something faster than (mapcar (lambda (x) ...) seq)?

yyoncho:
> In that case we would need another dap-mode backend, e.g. in
> dap-configuration-snippets.el, and we would need to add more variables. I will
> implement that.

Lets see how it goes, but it seems like it should go in dap-debug-template-configurations.

> Is there a way to get Emacs' JSON parser to include line number information?
> Because the thought of writing many jq (for the line numbers, goto launch
> configuration, feature 1.) scripts and elisp adapters to support lots of
> different launch configuration formats does not excite me.

I am not aware of a package providing this functionality. Maybe tree-sitter has such capability, but it we can have it only as optional dependency. The alternative is to write some rudimentary parser. Check also what lsp-json provides. We might check how VScode does that.

nbfalcon:
Is there some way to override the lsp workspace?

yyoncho:
> Is there a function to do mapcar, but with a body instead of a function? I.e. is there something faster than (mapcar (lambda (x) ...) seq)?

Consider asking those questions in gitter. There is --map, dolist, loop.

ericdallo:
Feel free to ask here: https://gitter.im/emacs-lsp/lsp-mode

nbfalcon:
Ok, I will ask in Gitter next time. Meanwhile I have written the patch to support package.json launch configurations, which you may review now (current master of my fork). One thing that will be a problem is how to choose the debugger to be used. In the package.json that @yyoncho linked, there were two sets of launch configurations for two debuggers in "contributes/debuggers/*/configurationSnippets/*". The problem is that they can share names, which would result in rather weird output. To handle this, I added the function dap-package-select-debugger, which currently only returns the first one. This should be improved, and perhaps made user customizable?

yyoncho:
> This should be improved, and perhaps made user customizable?

Most likely it should be part of `dap-utils-vscode-setup-function`.

nbfalcon:
I do not see how that function is relevant? Maybe my wording was off: if you look at https://github.com/golang/vscode-go/blob/master/package.json#L375, you will see that there is a "debuggers" key there. It is in fact an array, and there is an entry for every supported debugger (in this case: "go" and "godlvdap"). Both of these entries have the same set of launch configuration names. There however needs to be a way to choose which one to use.

yyoncho:
This function configures the location of the vscode package, it generates the downloader, etc - thus, it makes sense to configure the path to the debugger configurations.

nbfalcon:
I now understand what "configurationSnippets" in package.json is. I propose the following: we could write a shell script that converts package.json files to a directory of yasnippets, one per debugger. There could be a function "dap-launch-add-configuration" which would ask the user for a debugger and a template, and that would pop up a json buffer with the corresponding yasnippet already in place. When the user presses something like C-c C-c, we could take this json buffer and add it to the user's launch.json.

This feature requires yasnippet, but it can be optional and as such there would be no required dependency on yas.
The conversion from package.json to yasnippet templates could be done in a github action, and as such there shouldn't be any additional dependencies from that.

nbfalcon:
I have been looking into JSON schemas and it appears that they are suitable for implementing auto-complete in launch.json files, and it seems that this is the way VSCode does it, too. Someone would have to write a schema, and it would have to be added to lsp-json-schemas.

nbfalcon:
Lastly, launch.json goto configuration can be rather easily implemented with the help of jq (an optional feature, since not everyone wants jq): before parsing the launch.json file, make jq add location information to each object, in the form of a "dap-mode-context" key. E.g.
```json
{
    "configurations": [
        {
            "name": "....",
            "type": "python"
        }
    ]
}
```
becomes
```json
{
    "configurations": [
        {
            "dap-mode-context": {
                "file": "launch.json",
                "start": "3:8",
                "ends": "7:4"
            },
            "name": "....",
            "type": "python"
        }
    ]
}
```
Each configuration provider could optionally provide "dap-mode-context" and then that configuration could be opened easily in a helm menu. That key would then be stripped in dap-start-debugging. The key would also be present in each entry in the launch configuration history, so a function like "dap-goto-configuration" could use the dap-debug history.

nbfalcon:
Goto configuration seems hard to implement: jq's input_line_number is not helpful, as it just prints the line number of the last object parsed (which would be the entire file). I failed to implement it even with streaming parsing. Writing a JSON parser just for this feature seems inelegant.

nbfalcon:
@yyoncho I have written a launch.json schema generator, implementing your feature request 2., which could be used to provide auto-complete in launch.json files (that is, I tested lsp-mode jsonls auto-complete in a launch.json file) The project is here: https://github.com/nbfalcon/launch-json-schema. It can derive a launch.json schema from the package.json files of one or more VSCode debug extensions. One issue I could see is with copyright: basically, a derived work from multiple projects is created and the result shipped. The required license is unclear. As for features 1. and 2., I don't think I will implement them any time soon. They seem like a lot of work for a rather small gain.

yyoncho:
I checked the code briefly - can we do the schema generation in elisp? I do not like the idea of having external dependency for that transformation. Is 3 really a lot of work?

nbfalcon:
We propably could do that in elisp, but what would be the gain? The schema generation wouldn't be done on dap-user's computers. Rather, we would generate it in some github action and store the result. lsp-json-schemas could then contain an entry ("launch.json" . link-to-github-blobstore-with-generate-launch.json-schema).

nbfalcon:
Doing it in elisp would just replace jq with emacs, and emacs is much heavier.

nbfalcon:
Regarding 3.: well, it is at least as much work as 2.: another jq script would have to be written to extract the schemas. Then, the variable expansion code's launch configuration walker would have to be refactored to be general, and some more variable expansion code would have to be written. I would say the complexity is about 50 lines of elisp + 20 lines of jq, but note that I am not that good at estimating. The gain is however small, since especially with launch.json schema auto complete writing launch configurations manually wouldn't be that much work. If you really want that feature, I could look into it.

yyoncho:
> We propably could do that in elisp, but what would be the gain?

It will be a self-contained solution. Performance is not an issue unless there is a perf report proving the opposite. In this case, 
I would speculate that parsing 1mb json + generating the schema will most likely fit in 50ms or so. 

> If you really want that feature, I could look into it.

It is not up to me - you could pick whatever task you want. We are generally aiming for feature parity with vscode so everything that vscode do is kind of "in scope".

nbfalcon:
Okay, I will also write an elisp version. This won't be a large effort, since the "problems" (schema) are already solved in the jq scripts.

shackra:
I pulled 99fd1ed2e3e7c6e9ccaa3d0c61481885ae4bf52e and build the package using straight.el, put a `launch.json` file on the root of my project and after doing `dap-debug` the list of options available does not show any option named after the name of the debugging configuration of my launch.json file.

Any idea?

nbfalcon:
Works for me, with the latest dap-mode (eval-buffer). Can you please tell me the value of `dap-launch-configuration-providers'? Can you also perhaps attach your launch configuration?

nbfalcon:
Just C-h v dap-launch-configuration-providers.

nbfalcon:
Another issue I could see is with your lsp-workspace-root: please evaluate (lsp-workspace-root) (you can use M-: for that) and make sure it points to your real project root. This is where the launch.json file is looked up.

shackra:
> Just C-h v dap-launch-configuration-providers.

```
(dap-debug-template-configuration-provider dap-launch-find-parse-launch-json)
```

```
(lsp-workspace-root)
```

correctly points at the root of my project

yyoncho:
> Another issue I could see is with your lsp-workspace-root: please evaluate (lsp-workspace-root) (you can use M-: for that) and make sure it points to your real project root. This is where the launch.json file is looked up.

We should add log messages when where we are looking for this launch json, etc. 


@shackra can you post your launch.json?

shackra:
> Can you also perhaps attach your launch configuration?

the file at `(concat (lip-workspace-root) "/launch.json")`:

```
{
  "version": "0.2.0",
  "configurations": [
    {
      "name": "Debug pkg/request tests",
      "type": "go",
      "request": "launch",
      "mode": "test",
      "program": "${workspaceRoot}/pkg/request",
      "port": 2345,
      "host": "127.0.0.1",
      "remotePath": "",
      "showLog": true
    }
  ]
}
```

nbfalcon:
Well, one issue I can see is that you use ${workspaceRoot}: that is wrong, that variable does not exist. Use ${workspaceFolder} instead. There are currently some issues with variable expansion that I am fixing, so perhaps try changing that?. Still, it should show up.

nbfalcon:
Please evaluate (dap-launch-get-launch-json). Does it give you some JSON tree, or does it throw "Opening input file: No such file or directory"?

shackra:
I get back 

```
(:version "0.2.0" :configurations ((:name "Debug pkg/request tests" :type "go" :request "launch" :mode "test" :program "${workspaceRoot}/pkg/request" :port 2345 ...)))
```

nbfalcon:
You are running dap-debug, right? dap-debug-edit-template does not support launch.json atm. Please post the value of dap-launch-configuration-providers.

shackra:
> You are running dap-debug, right? dap-debug-edit-template does not support launch.json atm. Please post the value of dap-launch-configuration-providers.

```
(dap-debug-template-configurations-provider dap-launch-find-parse-launch-json)
```

nbfalcon:
Try replacing ${workspaceRoot} with ${workspaceFolder}.

shackra:
> Try replacing ${workspaceRoot} with ${workspaceFolder}.

```
(:version "0.2.0" :configurations ((:name "Debug pkg/request tests" :type "go" :request "launch" :mode "test" :program "${workspaceFolder}/pkg/request" :port 2345 ...)))
```

nbfalcon:
Does it work? Try evaluating (-mapcat #'funcall dap-launch-configuration-providers) with your go source file buffer current (important).

shackra:
> Does it work? 

sadly, no, I still see no `Debug pkg/request tests` in the select.

> Try evaluating (-mapcat #'funcall dap-launch-configuration-providers) with your go source file buffer current (important).

Got this:

```
((Go Connect Remote dlv Configuration :type go :request launch :name Connect to Remote dlv :mode remote :program nil :args nil :env nil :envFile nil) (Go Attach Executable Configuration :type go :request attach :name Attach to Executable :mode local :program nil :args nil :env nil :envFile nil) (Go Launch Executable Configuration :type go :request launch :name Launch Executable :mode exec :program nil :args nil :env nil :envFile nil) (Go Launch Unoptimized Debug Package Configuration :type go :request launch :name Launch Unoptimized Debug Package :mode debug :program nil :buildFlags -gcflags '-N -l' :args nil :env nil :envFile nil) (Go Launch Debug Package Configuration :type go :request launch :name Launch Debug Package :mode debug :program nil :buildFlags nil :args nil :env nil :envFile nil) (Go Launch File Configuration :type go :request launch :name Launch File :mode auto :program nil :buildFlags nil :args nil :env nil :envFile nil) (Python :: Run pytest (at point) :type python-test-at-point :args  :module pytest :request launch :name Python :: Run pytest (at point)) (Python :: Run pytest (buffer) :type python :args  :cwd nil :program nil :module pytest :request launch :name Python :: Run pytest (buffer)) (Python :: Run file (buffer) :type python :args  :cwd nil :module nil :program nil :request launch :name Python :: Run file (buffer)) (Chrome Browse URL :type chrome :cwd nil :mode url :request launch :webRoot nil :url nil :name Chrome Browse URL) (Chrome Browse File :type chrome :mode file :cwd nil :request launch :file nil :reAttach t :name Chrome Browse File) (Firefox Run Configuration :type firefox :cwd nil :request launch :file nil :reAttach t :program nil :name Firefox::Run) (Debug pkg/request tests :name Debug pkg/request tests :type go :request launch :mode test :program ${workspaceFolder}/pkg/request :port 2345 :host 127.0.0.1 :remotePath  :showLog t))
```

looks like it appears at the end

shackra:
despite that, `dap-debug` shows the following:

Captura de Pantalla 2020-08-11 a la(s) 14 34 37

shackra:
oh, nvm, I can see it below "Chrome Browse File"

nbfalcon:
Everything working then?

shackra:
yes

nbfalcon:
Great :)

Issue #329, opened 2020-08-10
pls fix read watch exression syntax error


wztdream:
Hi,
Here is a syntax error when read a string to watch the expression

dap-ui.el
around line 873

(read-string) receive string not symbol

`                       (t (symbol-name (symbol-at-point)))))))
`

ericdallo:
Thanks for the report, do you mind opening a PR? then other people can review the changes :)

wztdream:
ok

wztdream:
PR #330

Issue #328, opened 2020-08-10
Pls also update "watch expression"  buffer format to show array in one line


wztdream:
Hi,
Similar with [#322 ](https://github.com/emacs-lsp/dap-mode/commit/8c27a6fbe843e60308afda24277bae2cc39966bc), pls also fix "watch expression buffer" format to show variable in one line.

Maybe there is other locations need to update?

dap-ui.el
around line: 920
```
                             :label ,(concat (propertize (format "%s: " expression) 'face 'font-lock-variable-name-face)
                                             (propertize (s-truncate dap-ui-variable-length
                                                                     (s-replace "\n" "\\n" result))
                                                         'help-echo result))

```

wztdream:
PR #330

PR #330, opened 2020-08-10
fix small issues

wztdream:
1. fix read string syntax error #329 
2. show variable in Expression buffer in one line #328


yyoncho:
Thank you for contributing to `dap-mode` and congrats for our first `dap-mode` PR! :tada:

PR #333, opened 2020-08-11
Add escaping; stricter variable regexes; fix env: bug

nbfalcon:
Variable blocks ${...} can now be escaped. \${...} always becomes ${...}.
Additionally, '}' can now be escaped in variable blocks. Previously, there was
no way to refer to a 'env:}' variable, for example. Now, this can be done with
${env:\}}.

All variable-related regular expressions now require that the entire variable
match. Previously, ${workspaceFolderFooBar} was equivalent to
${workspaceFolder}. Fix this, throwing an error in that case.

The ${env:...} variable had a bug causing a strange error to be thrown if the
environment variable references did not exist. Fix this, making it warn and
return the empty string instead.

There is still more to come in this feature branch, including support for numbered prompting variables. However, this commit contains a bug-fix, so you may want to merge it now?

### DIFF
--- a/dap-variables.el
+++ b/dap-variables.el
@@ -229,7 +229,7 @@ back to nil.")
      numbered-prompts)
     extra-vars))
 
-(defun dap-variables--do-prompts-reset (_)
+(defun dap-variables--do-prompts-reset (&optional _)
   "Ask and reset the questions in `dap-variables-numbered-prompts'.
 For details, see `dap-variables--do-prompts'."
   (unwind-protect (dap-variables--do-prompts)
--- a/dap-variables.el
+++ b/dap-variables.el
@@ -150,8 +150,7 @@ and it must be sorted by NUMBER."
         (setq count (1+ count))))
     count))
 
-(defvar dap-variables-pre-expand-hook
-  '((lambda (_) (setq dap-variables-numbered-prompts '())))
+(defvar dap-variables-pre-expand-hook '()
   "List of functions to be run before a launch configuration is expanded.
 They take one argument: the run configuration.")
 
@@ -190,16 +189,16 @@ history argument. puthash this variable under the prompt back
 into `dap-variables--prompt-histories' and then finally setq this
 back to nil.")
 
-(defun dap-variables--do-prompts (_)
+(defun dap-variables--do-prompts ()
   "Ask the questions in `dap-variables-numbered-prompts' in correct order."
   (let* ((prev-id nil)
-        (prev-answer nil)
-        (extra-vars '())
-        (current-promptn 1)
-        (numbered-prompts (--sort (< (car it) (car other))
-                                  dap-variables-numbered-prompts))
-        (unique-prompts (dap-variables-count-unique-numbered-prompts
-                                  numbered-prompts)))
+         (prev-answer nil)
+         (extra-vars '())
+         (current-promptn 1)
+         (numbered-prompts (--sort (< (car it) (car other))
+                                   dap-variables-numbered-prompts))
+         (unique-prompts (dap-variables-count-unique-numbered-prompts
+                          numbered-prompts)))
     (mapc
      (-lambda ((id prompt var))
        (if (eq prev-id id) ;; prev-id can be nil, so eq and not =
@@ -230,9 +229,14 @@ back to nil.")
      numbered-prompts)
     extra-vars))
 
+(defun dap-variables--do-prompts-reset (_)
+  "Ask and reset the questions in `dap-variables-numbered-prompts'.
+For details, see `dap-variables--do-prompts'."
+  (unwind-protect (dap-variables--do-prompts)
+    (setq dap-variables-numbered-prompts nil)))
+
 (defvar dap-variables-post-walk-hook
-  '(dap-variables--do-prompts
-    (lambda (_) (setq dap-variables-numbered-prompts '())))
+  '(dap-variables--do-prompts-reset)
   "Functions to be run after first walking the launch configuration.
 When expanding a launch configuration, first
 `dap-variables-pre-expand-hook' is called. Then, the launch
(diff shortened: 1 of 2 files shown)


dap-variables.el
```diff
@@ -164,11 +178,15 @@ Return the result."
       (goto-char (point-min))
 
       (save-match-data
-        (while (re-search-forward "${\\([^}]*\\)}" nil t)
-          (let ((var (match-string 1)))
-            (replace-match
-             (with-current-buffer old-buffer
-               (dap-variables-expand-variable var))))))
+        (while (re-search-forward
+                "\\(\\\\\\$\\)\\|${\\(\\([^}\\]\\|\\(\\\\.\\)\\)*\\)}" nil t)
```
ericdallo, line 182: Could you move this regex to a defconst with a nice name?

kiennq: I would recommend to use `rx` for better maintenance

nbfalcon: I would love to use rx. Are we allowed to do that, as it is a dependency?

kiennq: `rx` is built-in. isn't it?

nbfalcon:
The second commit is rather large, implementing an architecture overhaul. Please see the commit message for details.
What is left to be done for the future:
- Allow custom ${NUMBER:PROMPT} with lambdas, to allow more advanced behaviour, for example: remote pid auto-complete
- Per PROMPT history
- regex prompts (see point 1)
- variable expansion in variables (for example: "${2: Enter your password for host '${1: Enter remote host}'}
All but the last feature should be easy to implement, without major architectural changes. The main bonus this change gives us, however, is the ability to implement more-advanced templates, as can be found in package.json files of VSCode debug plugins, most of which leverage variables in the form I just implemented. With some transformations, package.json templates could perhaps be made usable, even without modifying DAP as it is after this change.

nbfalcon:
What I think users will like most is the ability to make their launch configurations more dynamic: PIDs of "target" blocks no longer need to be hardcoded and changed every time a program is run.

nbfalcon:
The last commit (implement per prompting variable histories) has the following issue: the temporary variable hack (dap-variables--temp-hist) would be saved by savehist. Is there perhaps some way we could mitigate this? And the variable dap-variables-prompt-histories should be saved by savehist. How should we implement this? Should I add (push 'dap-variables-prompt-histories savehist-additional-variables) in dap-variables.el? Perhaps in the spacemacs layer? Also, should we make per variable histories optional?

dap-variables.el
```diff
@@ -138,50 +258,114 @@ Otherwise, return VALUE"
...
+  (cdr (cl-find-if (lambda (x) (string-match (car x) var)) variable-alist)))
+
+(defun dap-variables-expand-alist-variable (var variable-alist)
+  "Expand VAR by looking it up in VARIABLE-ALIST."
+  (save-match-data
+    (if-let ((value (dap-variables-find-matching var variable-alist)))
+        (or (dap-variables--eval-poly-type
+             value (if (= (length (match-data)) 2) nil var))
+            (progn (lsp-warn "launch.json: variable ${%s} is nil here" var) ""))
+      (progn (lsp-warn "launch.json: variable ${%s} is unknown" var) ""))))
```
yyoncho, line 276: no need progn in else block.

nbfalcon: Fixed.

nbfalcon:
One inelegancy still remains: the way I implemented per-prompt histories is with the hack described in dap-variables--temp-hist. Since this is passed to read-string, it is added to the list of variables that savehist saves. This causes a race condition: before it can be setq'd back to nil, it could already have been written to file. But I don't see any way how we could prevent save-hist from saving that variable, and there seems to be no other way to implement that feature.

yyoncho:
IMO the simplest option is to use one variable for all prompts. 

There are other options e. g.(disable savehist for the running of the prompt, generate unique variables runtime and drop dap-variables-prompt-his) but they will make the code more complex.

nbfalcon:
Also, can we make dap-launch-find-parse-launch-json the first provider in 'dap-launch-configuration-providers'? I use only launch.json files, so the configurations in dap-debug-template-configurations don't interest me, and perhaps that is the same for most other users.

yyoncho:
> and perhaps that is the same for most other users.

why do you think so considering that dap-mode is running for like 2 years and I think that launch.json was requested once or twice?

nbfalcon:
> > and perhaps that is the same for most other users.
> 
> why do you think so considering that dap-mode is running for like 2 years and I think that launch.json was requested once or twice?

Well, the launch.json provider does not generate anything if there is no launch.json file, so the users that don't want it wouldn't see anything, and the ones that want it would see their launch.json results first. I'll just use the following snippet in my dotspacemacs/user-init for now:
```elisp
(setq dap-launch-configuration-providers
      '(dap-launch-find-parse-launch-json
        dap-debug-template-configurations-provider))
```

yyoncho:
That make sense - move the JSON one on first position

nbfalcon:
dap-variables-numered-prompts might not be set to nil if the user cancels before debugging starts. This would not cause any issues, but it would be a minor waste of space. Is there perhaps some thing as a weak global?

yyoncho:
> dap-variables-numered-prompts might not be set to nil if the user cancels before debugging starts. This would not cause any issues, but it would be a minor waste of space. Is there perhaps some thing as a weak global?

I guess it is not a lot of data? Also, isn't unwind-protect sufficient to fix that?

nbfalcon:
> > dap-variables-numered-prompts might not be set to nil if the user cancels before debugging starts. This would not cause any issues, but it would be a minor waste of space. Is there perhaps some thing as a weak global?
> 
> I guess it is not a lot of data?

No its not. It depends on how many variable prompts ${1:host?} the user has, but I doubt anyone would have launch configuration/template with 20 prompts, ever. So, a list of max 20 elements. It would be cleared the next time the user runs dap-debug.

> Also, isn't unwind-protect sufficient to fix that?
I didn't know about that - I will look into it.

nbfalcon:
I use unwind-protect now. This makes for more elegant code, and prevents conses from being leaked in dap-variables-numbered-prompts.

yyoncho:
can you resolve the conflict?

nbfalcon:
Fixes in latest master.

yyoncho:
Thank you!

nbfalcon:
Great!

Issue #335, opened 2020-08-13
dap-debug not work | related with launch.json


rafamedina:
Hi

The `dap-debug` command is returning this error:
version:   ##dap-mode-20200811.436

`helm-M-x-execute-command: Opening input file: No such file or directory, /Users/rafael.medina/dev/source/fluttertesting/hello_flutter/launch.json`

Related
#https://github.com/emacs-lsp/lsp-dart/issues/52

ericdallo:
@nbfalcon could you take a look?

yyoncho:
@rafamedina workaround:

```
(setq dap-launch-configuration-providers  '(dap-debug-template-configurations-provider))
```

nbfalcon:
What a blunder on my side: I should check if the file exists before returning the filename in dap-launch-find-launch-json. I tested it on projects without launch.json files, but they weren't lsp projects. I guess I overlooked that.
At line 30:
dap-launch-find-launch-json should be:
```elisp
(defun dap-launch-find-launch-json ()
  "Return the location of the launch.json file in the current project."
  (when-let ((project (lsp-workspace-root))
             (launch-json (concat project (file-name-as-directory "launch.json"))))
    (when (file-exists-p launch-json)
      launch-json)))
```

nbfalcon:
@yyoncho that is a temporary fix. The launch.json provider shouldn't fail if the project has no launch.json file.

nbfalcon:
I'm making a PR against master asap.

nbfalcon:
https://github.com/emacs-lsp/dap-mode/pull/336

ericdallo:
It should be fixed on master now @rafamedina and soon on melpa latest release

nbfalcon:
Sorry for the inconvenience.

ericdallo:
No problem @nbfalcon, thanks for the quick fix :)

PR #336, opened 2020-08-13
In dap-launch-find-launch-json, check if it exists before returning

nbfalcon:
Fixes issue (#335).

### DIFF
--- a/dap-launch.el
+++ b/dap-launch.el
@@ -29,8 +29,10 @@
 
 (defun dap-launch-find-launch-json ()
   "Return the location of the launch.json file in the current project."
-  (when-let ((project (lsp-workspace-root)))
-    (concat project "/launch.json")))
+  (when-let ((project (lsp-workspace-root))
+             (launch-json (concat project (file-name-as-directory "launch.json"))))
+    (when (file-exists-p launch-json)
+      launch-json)))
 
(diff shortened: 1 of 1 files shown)


nbfalcon:
I got the issue wrong in the commit-message, though :(

nbfalcon:
Actually, wait, dont merge! It doesnt work.

dap-launch.el
```diff
@@ -29,8 +29,10 @@
 
 (defun dap-launch-find-launch-json ()
   "Return the location of the launch.json file in the current project."
-  (when-let ((project (lsp-workspace-root)))
-    (concat project "/launch.json")))
+  (when-let ((project (lsp-workspace-root))
```
ericdallo, line 32: I think this should be:
```suggestion
  (-when-let* ((project (lsp-workspace-root))
```

nbfalcon: No. when-let allows that. If you dig trough its source, you will find a reference to if-let*, being defined as (let*).

nbfalcon:
The last one works. I tested it by renaming my launch.json, after which its configurations didn't show up in dap-debug, and without renaming it. When not renaming it, its configuration showed up correctly.

ericdallo:
Thank you!

PR #349, opened 2020-08-28
Fix watch expression when using with code-debug

wangnangg:
watch expression doesn't work with code-debug. Looking at its [source code](https://github.com/WebFreak001/code-debug/blob/f2923480e45874324ca94badbe35c7ed80a5e172/src/mibase.ts#L622), it seems code-debug expects evaluateRequest's arg.context to be either "watch" or 'hover". Otherwise, it will assume "repl" context and pass the expression as is to gdb, resulting in error. I don't know who's at fault here. Anyway, here's the change that would make watch expression work.

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -1151,7 +1151,8 @@ RESULT to use for the callback."
         (dap--send-message
          (dap--make-request "evaluate"
                             (list :expression expression
-                                  :frameId active-frame-id))
+                                  :frameId active-frame-id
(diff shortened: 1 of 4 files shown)


yyoncho:
Looks good, willing to fix the other invocations with the proper `context`?

wangnangg:
yeah, sure. Updated the patch. please Check.

yyoncho:
Thank you!

Issue #350, opened 2020-08-28
Native/C++ Debug produces "Error: write EPIPE when starting a dap session"


mark-poscablo:
Mac OS Catalina 10.15.6
Spacemacs develop at `15d93914f`
dap-mode installed through MELPA, located at `/Users/mark/.emacs.d/elpa/26.3/develop/dap-mode-20200828.629`
Extension located at `/Users/mark/.vscode/extensions/webfreak.debug-0.25.0`
These lines are inside my `dotspacemacs/user-config`:
```elisp
  (setq dap-gdb-lldb-path "/Users/mark/.vscode/extensions/webfreak.debug-0.25.0")
  (setq dap-gdb-lldb-debug-program `("node" , "/Users/mark/.vscode/extensions/webfreak.debug-0.25.0/out/src/gdb.js"))
  (setq dap-gdb-lldb-path-lldb `("node" , "/Users/mark/.vscode/extensions/webfreak.debug-0.25.0/out/src/lldb.js"))
```
My debug configuration looks like this:
```elisp
(dap-register-debug-template
  "debug-session"
  (list :type "lldb"
        :request "launch"
        :name "debug-session"
        :target "/Users/mark/devel/myprogram/build/a.out"
        :cwd "/Users/mark/devel/myprogram"))
```
Running `dap-debug` (`SPC m d d d`) will:
* Ask me to "Select configuration template". I choose the configuration named above (`debug-session`)
* Produce this output in the `*Messages*` buffer:
```
error in process filter: dap--send-message: Process debug-session not running
error in process filter: Process debug-session not running
Debug session process exited with status: exited abnormally with code 1
```
* Produce this output in the `debug-session stderr` buffer:
```
events.js:292
      throw er; // Unhandled 'error' event
      ^

Error: write EPIPE
    at afterWriteDispatched (internal/stream_base_commons.js:154:25)
    at writeGeneric (internal/stream_base_commons.js:145:3)
    at Socket._writeGeneric (net.js:782:11)
    at Socket._write (net.js:794:8)
    at writeOrBuffer (_stream_writable.js:352:12)
    at Socket.Writable.write (_stream_writable.js:303:10)
    at MI2_LLDB.sendRaw (/Users/mark/.vscode/extensions/webfreak.debug-0.25.0/out/src/backend/mi2/mi2.js:766:32)
    at /Users/mark/.vscode/extensions/webfreak.debug-0.25.0/out/src/backend/mi2/mi2.js:793:18
[...]

PR #351, opened 2020-08-28
Improve variable expansion; add tests for it (fixes #346)

### LINKED ISSUE #346: `dap-variables-expand-in-launch-configuration` appends nil when `:command` is a list of a single element
Steps to reproduce the problem:

Just eval this from the scratch buffer:

```
(dap-variables-expand-in-launch-configuration '(:name "Debug"
 :connection-type 'pipe
 :coding 'no-conversion
 :command '("lldb-vscode")
 :stderr "*Whatever stderr*"
 :noquery t))
```

Note that the resulting form has the following `:command`:

```
[...]
:command '("lldb-vscode" nil)
[...]
```

This nil element causes an error later on when a call to `make-process` is built and run.

nbfalcon:
Handle lists generically in `dap-variables-walk-launch-configuration':
previously, alists were detected and handled with -all? #'consp, which is not
ideal, as it precludes lists of lists from being expanded correctly; however,
worse than that, all other types of lists were automatically assumed to be
plists. This led to a bug if a list with an odd number of elements was to be
expanded: it would end in nil. Fix this by considering any list a normal list,
and by handling conses whose `cdr' is not a list specially instead.

Also support vectors correctly now (yield a vector if a vector is given).

Test the variable expansion code in tests/dap-variables-test.el, with a set of
tests exercising previous issues.

### DIFF
--- a/dap-variables.el
+++ b/dap-variables.el
@@ -268,11 +268,11 @@ Otherwise, return VALUE"
         (t value)))
 
 (defun dap-variables-find-matching (var variable-alist)
-  "Return the VALUE whose REGEX matches VAR, or nil.
-VARIABLE-ALIST is a list of the form (REGEX . VARIABLE). This
-function modifies the `match-data'. REGEX may contain captures,
-which may be accessed with regular (`match-string' <n> VAR) or
-`match-data'."
+  "Return the corresponding VALUE to the REGEX matching VAR.
+Return nil if no matching VALUE is found. VARIABLE-ALIST is a
+list of the form (REGEX . VARIABLE). This function modifies the
+`match-data'. REGEX may contain captures, which may be accessed
+with regular (`match-string' <n> VAR) or `match-data'."
   (cdr (cl-find-if (lambda (x) (string-match (car x) var)) variable-alist)))
 
 (defun dap-variables-expand-alist-variable (var variable-alist)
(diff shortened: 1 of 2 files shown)


yyoncho:
Can you check if an element is a function and skip it?

nbfalcon:
Well, as of this commit a function would just get passed trough:
```elisp
(dap-variables-expand-in-launch-configuration (lambda ()))
```
yields (lambda nil).

nbfalcon:
Or do you mean to ignore it?

yyoncho:
Yes. When it is in value position.

yyoncho:
I had that fix in my local copy. Not sure where it was causing trouble.

nbfalcon:
What do you mean by value position? Can you show a list with some input to dap-variables-expand-in-launch-configuration and the expected result?

yyoncho:
`(list :startup-fn (lambda () "foo/bar"))`

nbfalcon:
Will yield (copied from *Messages* after eval-last-sexp):
```elisp
'(:startup-fn (lambda nil "foo/bar"))
```, which seems to be what you want?

**Comment by nbfalcon** [comment]
But I think only as of this branch.

**Comment by yyoncho** [comment]
Fine, can you add it as a test and it is good to go

**Comment by nbfalcon** [comment]
Consider also this snippet:
```elisp
(funcall
 (nth 1 (dap-variables-expand-in-launch-configuration
         (list :startup-fn (lambda () "foo/bar")))))
```
It will yield "foo/bar".

mark-poscablo:
@nbfalcon @yyoncho Thank you. I can verify that this also fixes the problem I described in #350.

nbfalcon:
Actually, your issue is not completely fixed yet. Unlike I thought, expansion in functions (lambdas, ...) is still performed (closures are lists). This will probably not cause any issues since lists are handled better now. However, if you have a lambda like (lambda () (message "${workspaceFolder}")) it would print your workspace folder instead. I will make a PR to address this soon, but I don't expect anyone to be affected by this issue.

You can see what I mean by evaluating the following snippet (in IELM, M-: or \*scratch*):
```elisp
(funcall (dap-variables-expand-in-launch-configuration
              (lambda () (message "${workspaceFolder}"))))
```

mark-poscablo:
Thanks @nbfalcon . I guess what I meant was that as far as my C++ debugging with dap-mode is concerned, the issue has been fixed - but of course I recognize that the case you just described is valid and can be a problem for other situations (or the same situation in the future as more modifications are made as well).

PR #352, opened 2020-08-29
add support for gdbserver

wangnangg:
Wonder if this will be useful for others. In embedded development environment, we usually use gdbserver to debug programs deployed in the target machine. This patch adds a configuration for gdbserver.


yyoncho:
Thank you!

PR #353, opened 2020-08-29
dap-variables-expand-in-launch-configuration: don't touch functions

nbfalcon:
Some types of function objects like closures are just lists of a special schema.
This means that variable expansion would be performed in them, which is
undesirable. Don't do that and verify the fix with a new test case.

See my latest comment on #346.


nbfalcon:
I have no idea why tests/dap-variables-test.el is binary. Does anyone know a fix?

PR #357, opened 2020-09-09
Simplify logic for py.

jcs090218:
This patch does the following.

* Fixed mix use of `space` and `tab`
* Simplify logic.


kiennq:
I think `dap-python` has some warning when byte compile, can you help resolving that too?

jcs090218:
I have never used `cl-defstruct` operation, is that how you solve the byte compile warnings? 😕 I should probably revert it if this doesn't work! Thanks!

yyoncho:
If you are willing to work on that we should revisit the idea of using defstructs here. It was supposed to be an intermediate state and evolve from that solution but that never happened.

yyoncho:
Also, there is `lsp--symbols->symbols-hierarchy` which should be usable by dap-python as well.

jcs090218:
> If you are willing to work on that we should revisit the idea of using defstructs here. It was supposed to be an intermediate state and evolve from that solution but that never happened.

Got it. If I am going to rework on this I might make another PR. Just want to keep each PR clean. :)

> Also, there is lsp--symbols->symbols-hierarchy which should be usable by dap-python as well.

No quite sure what you mean by this. 😕 Let me take some time and read through the code!

yyoncho:
> No quite sure what you mean by this.  Let me take some time and read through the code!

The code in dap-python tries to build a hierarchy to find out the current test method, test class, and so on. We do the same for the headerline. It will be nice if we can reuse the logic.

nbfalcon:
As a side note, you shouldn't end your commit headings in a period, and use the imperative mood in headings:
"Fixed byte compile warnings." -> "Fix byte compile warnings", ....
I found this guide to be a great resource myself: https://chris.beams.io/posts/git-commit/

PR #362, opened 2020-09-11
Implement :console "integratedTerminal"/"externalTerminal"

nbfalcon:
Note especially the breaking change to `dap--make-response'' api. I have grepped the codebase for references to that function and didn't find any, however there may be third party plugins (official?). The only place in the dap protocol where the *client* gets to make a response is with "runInTerminal".

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -78,7 +78,7 @@ should be used in `dap-internal-terminal-*'."
     (declare-function vterm-mode "vterm" (&optional arg))
     (let ((vterm-shell command)
           (vterm-kill-buffer-on-exit nil))
-      (vterm-mode 1)
+      (vterm-mode)
       ;; TODO: integrate into dap-ui
       (display-buffer (current-buffer)))))
 
@@ -90,7 +90,9 @@ should be used in `dap-internal-terminal-*'."
   "Run COMMAND with an auto-detected terminal.
 If `vterm' is loaded or auto-loaded, use vterm. Otherwise, use
 `async-shell-command'."
-  (if (fboundp 'vterm-mode)
+  ;; NOTE: 'vterm is autoloaded. This means that (fboundp 'vterm) will yield t
+  ;; even before vterm is loaded.
+  (if (fboundp 'vterm)
       (dap-internal-terminal-vterm command title debug-session)
     (dap-internal-terminal-shell command title debug-session)))
 
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -988,7 +988,9 @@ PARAMS are the event params.")
      (when (string= kind "external")
        (let* ((name (or title (concat (dap--debug-session-name debug-session)
                                       "- terminal"))))
-         (with-demoted-errors "dap-debug: failed to start external terminal: %s"
+         (with-demoted-errors "dap-debug: failed to start \
+external terminal: %s. Set `dap-external-terminal' to the correct
+value or install the terminal configured (probably xterm)."
            (apply #'start-process name name
                   (-map (lambda (part)
                           (->> part
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -48,7 +48,7 @@
   :group 'dap-mode
   :type 'boolean)
 
-(defcustom dap-external-terminal '("xterm" "-display" "{title}"
+(defcustom dap-external-terminal '("xterm" "-T" "{title}"
                                    "-hold" "-e" "sh" "-c" "exec {command}")
   "Command to launch the external terminal for debugging.
 When specifying that the program be run in an external terminal,
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -63,6 +63,54 @@ See also `dap-default-terminal-kind'."
   :group 'dap-mode
   :type '(repeat string))
 
+(defun dap--make-terminal-buffer (title debug-session)
+  "Generate an internal terminal buffer.
+The name is derived from TITLE and DEBUG-SESSION. This function
+should be used in `dap-internal-terminal-*'."
+  (generate-new-buffer
+   (format "*%s %s*"
+           (dap--debug-session-name debug-session)
+           (if title (concat "- " title) "console"))))
+
+(defun dap-internal-terminal-vterm (command title debug-session)
+  (with-current-buffer (dap--make-terminal-buffer title debug-session)
+    (require 'vterm)
+    (declare-function vterm-mode "vterm" (&optional arg))
+    (let ((vterm-shell command)
+          (vterm-kill-buffer-on-exit nil))
+      (vterm-mode 1)
+      ;; TODO: integrate into dap-ui
+      (display-buffer (current-buffer)))))
+
+(defun dap-internal-terminal-shell (command title debug-session)
+  (let ((buf (dap--make-terminal-buffer title debug-session)))
+    (async-shell-command command buf buf)))
+
+(defun dap-internal-terminal-auto (command title debug-session)
+  "Run COMMAND with an auto-detected terminal.
+If `vterm' is loaded or auto-loaded, use vterm. Otherwise, use
+`async-shell-command'."
+  (if (fboundp 'vterm-mode)
+      (dap-internal-terminal-vterm command title debug-session)
+    (dap-internal-terminal-shell command title debug-session)))
+
+(defcustom dap-internal-terminal #'dap-internal-terminal-auto
+  "Terminal used with :console \"integratedTerminal\".
+It is a function that shall take three arguments: the command to
+run, as a string, the title from the debug adapter (may be nil)
+and the debug session and it should execute COMMAND. Aside from
+that, it can do anything and is itself responsible for displaying
+any buffers, ....
+
+If you are looking at implementing your own such function, see
+also `dap--make-terminal-buffer'."
+  :group 'dap-mode
+  :type '(radio
+          (const :tag "auto-detect" :value dap-internal-terminal-auto)
+          (const :tag "vterm" :value dap-internal-terminal-vterm)
+          (const :tag "asnyc-shell" :value dap-internal-terminal-shell)
+          (function :tag "Custom function")))
+
 (defcustom dap-output-buffer-filter '("stdout" "stderr")
   "If non-nil, a list of output types to display in the debug output buffer."
   :group 'dap-mode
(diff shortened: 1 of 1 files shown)


nbfalcon:
@yyoncho: I made the internal terminal configurable, with the default to leverage vterm, if available, and to use async-shell-command instead. I tested both (manually).

nbfalcon:
Once this is merged, I'll add that feature to my "Acknowledgements" section, too.

nbfalcon:
@yyoncho We could have another variable `dap-server-start-function' and have it funcalled the same way. I could add another function like the others but using compilation-start. However, are you sure that compilation-start is not the right way to go about this? In the case of the "integratedTerminal", compilation-start would be the wrong solution, as compilation-start was not designed for interactive use. The case of starting servers isn't interactive though and errors should perhaps be highlighted. Grepping trough the code, I often see compilation-start being called with 'dap-server-log-mode together. What do you think?

yyoncho:
The issue with compilation-start is that it kills emacs when there are long lines in the output. In this case, it is better to lose some functionality, than unusably slow emacs. So even if the default is compilation-start it will be nice if we have the option to use vterm. As a side note, we might research if we can run the compilation mode over the vterm buffer. In this case, we will have the best of both worlds.

dap-mode.el
```diff
@@ -48,6 +48,69 @@
   :group 'dap-mode
   :type 'boolean)
 
+(defcustom dap-external-terminal '("xterm" "-display" "{title}"
```
yyoncho, line 51: can you change `-display` to `-T`?

dap-mode.el
```diff
@@ -48,6 +48,69 @@
...
+
+(defun dap-internal-terminal-shell (command title debug-session)
+  (let ((buf (dap--make-terminal-buffer title debug-session)))
+    (async-shell-command command buf buf)))
+
+(defun dap-internal-terminal-auto (command title debug-session)
+  "Run COMMAND with an auto-detected terminal.
+If `vterm' is loaded or auto-loaded, use vterm. Otherwise, use
+`async-shell-command'."
+  (if (fboundp 'vterm-mode)
```
yyoncho, line 93: use 'vterm instead of 'vterm-mode - vterm-mode is not autoloaded.

dap-mode.el
```diff
@@ -905,6 +968,45 @@ PARAMS are the event params.")
...
+                 :value "external"))
+  :group 'dap-mode)
+
+(defun dap--start-process (debug-session parsed-msg)
+  (-let* (((&hash "arguments" (&hash? "args" "cwd" "title" "kind") "seq")
+           parsed-msg)
+          (default-directory cwd)
+          (command-to-run (string-join args " ")))
+    (pcase (or kind dap-default-terminal-kind)
+      ("external"
```
yyoncho, line 985: external wont work unless xterm is defined. I suggest checking if xterm is present and if it is not present - show warning with instruction and use internal terminal. This is crucial for OOTB experience.

dap-mode.el
```diff
@@ -48,6 +48,69 @@
...
+  "Generate an internal terminal buffer.
+The name is derived from TITLE and DEBUG-SESSION. This function
+should be used in `dap-internal-terminal-*'."
+  (generate-new-buffer
+   (format "*%s %s*"
+           (dap--debug-session-name debug-session)
+           (if title (concat "- " title) "console"))))
+
+(defun dap-internal-terminal-vterm (command title debug-session)
+  (with-current-buffer (dap--make-terminal-buffer title debug-session)
```
yyoncho, line 76: With the code as it is with fairly recent spacemacs I get:

```
\\{vterm-mode-map}" (interactive) (progn (make-local-variable 'delay-mode-hooks) (let ((delay-mode-hooks t)) (kill-all-local-variables) (setq major-mode 'vterm-mode) (setq mode-name "VTerm") nil (use-local-map vterm-mode-map) (set-syntax-table vterm-mode-syntax-table) (setq local-abbrev-table vterm-mode-abbrev-table) (buffer-disable-undo) (and (boundp 'display-line-numbers) (let ((font-height (expt text-scale-mode-step text-scale-mode-amount))) (setq vterm--linenum-remapping (face-remap-add-relative 'line-number :height font-height)))) (let ((process-environment (append (cons (concat "TERM=" vterm-term-environment-variable) '("INSIDE_EMACS=vterm" "LINES" "COLUMNS")) process-environment)) (process-adaptive-read-buffering nil) (width (max (- (window-body-width) (vterm--get-margin-width)) vterm-min-window-width))) (setq vterm--term (vterm--new (window-body-height) width vterm-max-scrollback vterm-disable-bold-font vterm-disable-underline vterm-disable-inverse-video)) (setq buffer-read-only t) (set (make-local-variable 'scroll-conservatively) 101) (set (make-local-variable 'scroll-margin) 0) (set (make-local-variable 'hscroll-margin) 0) (set (make-local-variable 'hscroll-step) 1) (set (make-local-variable 'truncate-lines) t) (advice--add-function :filter-return (cons #'(lambda nil (advice--buffer-local 'filter-buffer-substring-function)) #'(lambda (gv--val) (advice--set-buffer-local 'filter-buffer-substring-function gv--val))) #'vterm--filter-buffer-substring nil) (setq vterm--process (make-process :name "vterm" :buffer (current-buffer) :command (list "/bin/sh" "-c" (format "stty -nl sane iutf8 erase ^? rows %d columns %d >/dev/null && exec %s" (window-body-height) width vterm-shell)) :coding 'no-conversion :connection-type 'pty :filter #'vterm--filter :sentinel (if (or vterm-exit-functions vterm-kill-buffer-on-exit) (progn #'vterm--sentinel))))) (add-hook 'change-major-mode-hook #'(lambda nil (interactive) (user-error "You cannot change major mode in vterm buffers")) nil t) (vterm--set-pty-name vterm--term (process-tty-name vterm--process)) (process-put vterm--process 'adjust-window-size-function #'vterm--window-adjust-process-window-size) (setq next-error-function 'vterm-next-error-function))) (run-mode-hooks 'vterm-mode-hook)), 1
error in process filter: Wrong number of arguments: ((vterm-mode-abbrev-table vterm-mode-syntax-table t) nil "Major mode for vterm buffer.
```

nbfalcon: I think it was a blunder related to (vterm-mode *1*). Derived modes don't take an ARG. Fixed (works with Spacemacs, I have the shell layer installed).

dap-mode.el
nbfalcon: Fixed. The entire block is wrapped in with-demoted-errors and will fall-back to using the built-in one instead. I tested with `dap-external-terminal' set to '("foobar") and with it set to my default KiTTY list, both working as expected. However, here I would like some more feedback: do you think with-demoted-errors is enough, or should I use condition-case + lsp--warn instead? Also, what do you think of the error message:
         (with-demoted-errors "dap-debug: failed to start \
external terminal: %s. Set `dap-external-terminal' to the correct
value or install the terminal configured (probably xterm)." ...)

yyoncho (approved):

dap-mode.el
yyoncho: I merged it as it is but I am a bit concerned here because there might be different errors coming from this block and the error message will be misleading.

nbfalcon: Well, all errors would be related to failure to start the terminal, which could be for a variety of reasons: terminal not installed/not in PATH, terminal *is* installed but permission is denied to launch it (SELinux, -x bit), which would handle all of that. Is your worry that dap--send-message might fail? I'll fix that. Aside from that, what do you propose?

yyoncho: I think it is good enough for now, lets see how it goes. A minor improvement would be to include the actual command(after replacement) that has failed in the error message.

nbfalcon: Would that even be possible using with-demoted-errors? The format string has to be static, doesn't it? Try evaluating the following in scratch:
```emacs-lisp
(let ((m "FooBAr%S"))
  (with-demoted-errors m
    (error "5")))
```
An error is still shown and there is no trace of m (as an error format) in macrostep.

nbfalcon: The fix to move dap--send-message outside of with-demoted-errors is in master.

yyoncho: > Would that even be possible using with-demoted-errors? 

with-demoted-errors is just a wrapper over `condition-case-unless-debug`

nbfalcon: Fixed in master. What I was afraid of is that the code could become horrible ugly (in terms of indent). Long string constants never look elegant....

dap-mode.el
```diff
@@ -1222,6 +1224,7 @@ DEBUG-SESSION is the active debug session."
         (run-hooks 'dap-breakpoints-changed-hook)))))
 
 (defun dap--breakpoint-filter-enabled (filter type default)
+  (defvar dap-exception-breakpoints) ;; NOTE: always called with dap-ui loaded
```
kiennq, line 1227: This should be moved to top level since we need this only for byte-compile. Just one time evaluation at top level is enough. You can see other file in this project is doing the same.

nbfalcon: I was waiting for that comment :). That defvar has no overhead. Its not even in the byte-code. Evaluate the following in scratch and you'll see what I mean:
```emacs-lisp
(disassemble (byte-compile '(lambda () (defvar foo) 500)))
```
-> constant 500; return
Defvar without a default-value is just a byte-compiler hint that it is a global variable and will be imported and let-bindings should always override. (Similar to declare-function).

dap-mode.el
```diff
@@ -1834,10 +1837,12 @@ point is set."
   :init-value nil
   :global t
   :group 'dap-mode
+  (declare-function dap-ui-mode "dap-ui" (&optional arg))
```
kiennq, line 1840: Same here, `declare-function` should be moved to top level (not inside any function/macro).
And actually `dap-ui-mode` is autoloaded so it will not require `declare-function`.
Only the `dap-ui-many-windows-mode` is require to be `declare-function`.

The rule of thumb is that unless byte-compile complains, we normally don't need `declare-function`.

nbfalcon: That is the point - there were byte-compile warnings. declare-funtion, like the defvar above, has no overhead (zero: not a byte of bytecode). It is a compile-time hint so that the compiler would not throw warnings.

dap-mode.el
kiennq: Well, I think that the problem here is the interpreted scope.
With `defvar` inside `defun` it will be understood as only available/effective inside that function.
If there's another one which also use this variable, it will need to `defvar` that variable too. No much different from `boundp` check.

Having all forward declaration in top level will prevent that, at least makes it clear that the variable is available in whole file scope already.

dap-mode.el
kiennq: Is there byte-compile warning for `dap-ui-mode`? Which is autoloaded?

dap-mode.el
nbfalcon: Well, one can always evil-search-forward? Having such declare-functions close to requires makes it clear what is actually imported, but what do you think?

dap-mode.el
nbfalcon: Yes. The byte compiler cannot know that some function, which is not loaded at global scope (that is, not even locally in a defun) be defined, and so will throw warnings. For such things, declare-function is made for: it tells the byte compiler that one is sure that some function will be available at run time, and optionally, what its signature is, eliminating such warnings.

dap-mode.el
kiennq: I think we can move the `defvar` to just a few line up, right before `defun`. In that way, it's still close to where it's first used + we will have it as top level declaration. And we don't have to write `defvar` or `declare-function` every time it being used in another function. To keep it consistent with the convention already used in this project.

nbfalcon: Ahh, if there is a convention I'll update it soon. I am however currently busy with the refactor to move dap-ui--show-buffer to display-buffer.

PR #138, opened 2019-10-15
Add preliminary netcoredbg support

TOTBWF:
This PR implements DAP support for projects on .NET core, using the [netcoredbg](https://github.com/Samsung/netcoredbg) debugger.

TODOS:
- [x] Attach to process
- [ ] Remote debugging
- [ ] Debugging inside of docker containers

PR #363, opened 2020-09-16
dap-netcore: Fix Windows install, and add "Attach to process"

sebasmonia:
These are the changes what lived in the comments of the original PR #138.

### DIFF
--- a/dap-netcore.el
+++ b/dap-netcore.el
@@ -79,11 +79,11 @@ The slash is expected at the end."
 (defun dap-netcore--populate-args (conf)
   "Populate CONF with arguments to launch or attach netcoredbg."
   (dap--put-if-absent conf :dap-server-path (list (dap-netcore--debugger-locate-or-install) "--interpreter=vscode"))
-  (setq conf (pcase (plist-get conf :mode)
-               ("launch"
-                (dap--put-if-absent conf :program (expand-file-name (read-file-name "Select an executable:"))))
-               ("attach"
-                (dap--put-if-absent conf :processId (string-to-number (read-string "Enter PID: " "2345")))))))
+  (pcase (plist-get conf :mode)
+    ("launch"
+     (dap--put-if-absent conf :program (expand-file-name (read-file-name "Select an executable:"))))
+    ("attach"
+     (dap--put-if-absent conf :processId (string-to-number (read-string "Enter PID: " "2345"))))))
 
 
 (dap-register-debug-provider
--- a/dap-netcore.el
+++ b/dap-netcore.el
@@ -48,7 +48,10 @@ The slash is expected at the end."
 
 (defun dap-netcore--debugger-install ()
   "Download the latest version of netcoredbg and extract it to `dap-netcore-install-dir'."
-  (let* ((temp-file (make-temp-file "netcoredbg" nil ".tar.gz"))
+  (let* ((temp-file (make-temp-file "netcoredbg" nil
+                                    (if (eq system-type 'windows-nt)
+                                        ".zip"
+                                      ".tar.gz")))
          (install-dir-full (expand-file-name dap-netcore-install-dir))
          (unzip-script (pcase system-type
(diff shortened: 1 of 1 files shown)


dap-netcore.el
```diff
@@ -73,20 +76,31 @@ The slash is expected at the end."
...
     dbg))
 
-(defun dap-netcore--populate-default-args (conf)
-  "Populate CONF with the default arguments."
-  (dap--put-if-absent conf :program (read-file-name "Select an executable:" (concat (lsp-workspace-root) "bin/Debug")))
-  (dap--put-if-absent conf :dap-server-path (list (dap-netcore--debugger-locate) "--interpreter=vscode")))
+(defun dap-netcore--populate-args (conf)
+  "Populate CONF with arguments to launch or attach netcoredbg."
+  (dap--put-if-absent conf :dap-server-path (list (dap-netcore--debugger-locate-or-install) "--interpreter=vscode"))
+  (setq conf (pcase (plist-get conf :mode)
```
yyoncho, line 82: nit: setq seems to be redundant.

yyoncho (approved):
Looks good to me from elisp perspective. 

I will keep it open for some time to allow someone to test it.

TOTBWF:
Just wanted to pop in to say thank you for continuing the work 🙂

yyoncho:
@sebasmonia thank you!

smotti:
Thanks a lot for the great work!

Though I've got some issues using it.
So I use a 64bit windows and Emacs 27.1. The installation of netcoredbg worked fine but what doesn't seem to work is setting breakpoints. I start a debug session by using the launch template. The application launches without issues and i can see the sessions threads (all no name). When I set breakpoints and than interact with the program to execute the code path where a breakpoint is set it just doesn't break and continues. Another thing is that the UI controls actually don't work. Meaning hitting the disconnect button does not disconnect (close) the session.

I'm fairly inexperienced when it comes to writing/debugging emacs packages so any guidance to help figure out what's going on is highly appreciated.

Thank you!

yyoncho:
> When I set breakpoints and than interact with the program to execute the code path where a breakpoint is set it just doesn't break and continues. 

I think that someone complained about that in another issue. I will need the help from someone using windows and that adapter so we can compare what vscode is sending with dap-mode

> Another thing is that the UI controls actually don't work.

You may open bug for that. Apparently, posframe library is tricky...

sebasmonia:
@smotti I had the same challenges, despite my best attempts I couldn't get netcoredbg to pick up the breakpoints.

I am using Windows less and less lately, and I have never used VS Code, as mentioned above, we need someone that does to help us fill in the gap.
Keep in mind VS Code _by default_ doesn't use netcoredbg so there is still a change the problem is in the debugger itself.

smotti:
@sebasmonia @yyoncho thanks for the quick reply.

I'm happy to help but might need some guidance on how to start digging deeper (though sebasmonia already provided some hints, but some general pointers at the emacs level will be appreciated). As I mentioned I'm fairly inexperienced writing & debugging emacs package.

One thing I found that actually could be the issue: https://github.com/Samsung/netcoredbg/issues/12

Because the path passed to netcoredbg is unix-style and all lowercase, though with the change addressed by the linked issue, the lowercase shouldn't matter but probably the usage of '/' compared to the windows '\'. At least this is the case for the dap-netcore package. I'll first look into replacing the slash with a backwards slash.

sebasmonia:
I tried to do that by manually editing editing the file where `dap-mode` keeps the breakpoint information, replacing the / by \\, but then the breakpoints weren't detected.
I guess advising some function in the dap-mode internals could do the trick?

smotti:
Yeah so i modified the `dap--breakpoints-changed` function to replace the slashes in `file-name`. But that didn't solve it. Currently actually trying out visual code with netcoredbg and there it also doesn't stop at breakpoints. Have to check if the change is actually in the version that was install.

PR #364, opened 2020-09-16
Add callback and version do download vscode extension util

ericdallo:
* Add version option
* Add callback to call after download


ericdallo:
@yyoncho lsp-dart will use vscode extension and not the github release after some extension changes from [here](https://github.com/emacs-lsp/lsp-dart/issues/55)

ericdallo:
I'm merging this to unblock `lsp-dart` and it looks simple, let me know if there's any problem

PR #373, opened 2020-09-27
[dap-lldb] Change :type to lldb-vscode, per the docs

leungbk:
The documentation says that the :type for both 'attach' and 'launch'
requests must be 'lldb-vscode'. (though it appears to work even when
the :type is wrong...)

An added benefit is that we no longer have any naming conflicts with
the dap-gdb-lldb's lldb debug adapter, which has :type 'lldb'.

----

#

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -2,6 +2,8 @@
 #+STARTUP: content
 
 * Changelog
+** 0.7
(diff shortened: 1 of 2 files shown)


yyoncho:
Looks good. Since this is a breaking change, can you drop a note in the changelog?

yyoncho:
@leungbk thank you. FYI do a ping after force push.

PR #376, opened 2020-09-29
Add a Makefile and fix byte-compilation warnings

leungbk:
----

#


dap-mode.el
```diff
@@ -1240,7 +1241,14 @@ DEBUG-SESSION is the active debug session."
              #'equal))
 
 (defun dap--set-exception-breakpoints (debug-session callback)
-  (-let [(&dap-session 'initialize-result 'launch-args (&plist :type)) debug-session]
+  ;; XXX: for whatever reason, make byte-compile emits warnings when
```
kiennq, line 1244: This seems weird, the production code has been proven working, perhaps we need to tune/require/shuffle something for this to work?

kiennq:
Do you want too update travis to use new make file or convert travis to Github CI?

dap-mode.el
yyoncho, line 1244: most likely the definition of the destructoring function should be in eval-when-compile block

dap-mouse.el
```diff
@@ -166,6 +167,8 @@ If there is an active selection - return it."
           (setq dap-mouse--hide-timer nil)
           (remove-hook 'post-command-hook #'dap-tooltip-post-tooltip)))))))
 
+(declare-function #'lsp-treemacs-render "ext:lsp-treemacs")
```
yyoncho, line 170: lsp-treemacs is already a dependency, so no need from declare-function just require it.

yyoncho (approved):

yyoncho:
Thank you!

PR #381, opened 2020-10-09
Strip trailing comments from launch.json

jrwats:
We already sanitize comments from JSON.  Let's also strip trailing commas.  I happen to have launch.json with trailing commas, which is out of my control at my workplace.  Furthermore, VSCode happily ingests these trailing newlines in the "JSON" launch configuration.  Emacs' DAP-mode should handle such "JSON" as well.

The regex works by finding any commas, followed any number of whitespace or newlines, followed by a closing brace or bracket.  The non-comma characters are placed in a separate regex group (2), so that they can replace the full match and only strip the comma itself.

With a file like the following
```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "type": "hhvm",
            "request": "attach",
            "name": "Attach to HHVM",
            "disableStdoutRedirection": false
        },
        {
            "type": "hhvm",
            "request": "launch",
            "name": "Launch Script",
            "launchScriptPath": "${file}",
        },
        {
            "type": "hhvm",
            "request": "launch",
            "name": "Launch Script Controller",
            "launchScriptPath": "${fileBasename}",
            "launchWrapperCommand": "${workspaceFolder}/scripts/init",
        },
	/* some silly comment */
    ],
}
```

We get the following diff
```
2,4c2,4
<     // Use IntelliSense to learn about possible attributes.
<     // Hover to view descriptions of existing attributes.
<     // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
---
>
>
>
17c17
<             "launchScriptPath": "${file}",
---
>             "launchScriptPath": "${file}"
24c24
<             "launchWrapperCommand": "${workspaceFolder}/scripts/init",
---
>             "launchWrapperCommand": "${workspaceFolder}/scripts/init"
26,27c26,27
< 	/* some silly comment */
<     ],
---
>
>     ]
```

resulting in
```
{



    "version": "0.2.0",
    "configurations": [
        {
            "type": "hhvm",
            "request": "attach",
            "name": "Attach to HHVM",
            "disableStdoutRedirection": false
        },
        {
            "type": "hhvm",
            "request": "launch",
            "name": "Launch Script",
            "launchScriptPath": "${file}"
        },
        {
            "type": "hhvm",
            "request": "launch",
            "name": "Launch Script Controller",
            "launchScriptPath": "${fileBasename}",
            "launchWrapperCommand": "${workspaceFolder}/scripts/init"
        },

    ]
}
```

### DIFF
--- a/dap-launch.el
+++ b/dap-launch.el
@@ -41,7 +41,7 @@ supported."
                 (or (: "//" (* nonl) eol)
                     (: "/*" (* (or (not (any ?*))
                                    (: (+ ?*) (not (any ?/))))) (+ ?*) ?/)
-                    (: "," (group (* (any blank ?\C-j)) (any ?\} ?\] )))))
+                    (: "," (group (* (any blank ?\C-j)) (any ?\} ?\])))))
                (: "\"" (* (or (not (any ?\\ ?\")) (: ?\\ nonl))) "\"")))
           nil t)
     ;; we matched a comment
--- a/dap-launch.el
+++ b/dap-launch.el
@@ -28,8 +28,8 @@
 (require 'f)
 (require 'rx)
 
-(defun dap-launch-remove-comments ()
-  "Remove all C-style comments in the current buffer.
+(defun dap-launch-sanitize-json()
+  "Remove all C-style comments and trailing commas in the current buffer.
 Comments in strings are ignored. The buffer is modified in place.
 Replacement starts at point, and strings before it are ignored,
 so you may want to move point to `point-min' with `goto-char'
@@ -40,12 +40,13 @@ supported."
            (or (group
                 (or (: "//" (* nonl) eol)
                     (: "/*" (* (or (not (any ?*))
-                                   (: (+ ?*) (not (any ?/))))) (+ ?*) ?/)))
+                                   (: (+ ?*) (not (any ?/))))) (+ ?*) ?/)
+                    (: "," (group (* (any blank ?\C-j)) (any ?\} ?\] )))))
                (: "\"" (* (or (not (any ?\\ ?\")) (: ?\\ nonl))) "\"")))
           nil t)
     ;; we matched a comment
     (when (match-beginning 1)
-      (replace-match ""))))
+      (replace-match (or (match-string 2) "")))))
 
 (defun dap-launch-find-launch-json ()
   "Return the location of the launch.json file in the current project."
@@ -66,7 +67,7 @@ supported."
     (with-temp-buffer
       ;; NOTE: insert-file-contents does not move point
       (insert-file-contents launch-json)
-      (dap-launch-remove-comments)
+      (dap-launch-sanitize-json)
       ;; dap-launch-remove-comments does move point
       (goto-char (point-min))
 
--- a/dap-launch.el
+++ b/dap-launch.el
@@ -28,7 +28,7 @@
 (require 'f)
 (require 'rx)
 
-(defun dap-launch-sanitize-json()
+(defun dap-launch-sanitize-json ()
   "Remove all C-style comments and trailing commas in the current buffer.
 Comments in strings are ignored. The buffer is modified in place.
 Replacement starts at point, and strings before it are ignored,
--- a/test/dap-test.el
+++ b/test/dap-test.el
@@ -123,13 +123,13 @@
     (should (dap-python--equal actual expected))
     (should (dap-python--equal (dap-python--nearest-test expected) "::TestClass::test_foo"))))
 
-(defun dap-launch-test--delete-string-comments (s)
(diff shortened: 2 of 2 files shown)


dap-launch.el
```diff
@@ -40,12 +40,13 @@ supported."
            (or (group
                 (or (: "//" (* nonl) eol)
                     (: "/*" (* (or (not (any ?*))
-                                   (: (+ ?*) (not (any ?/))))) (+ ?*) ?/)))
+                                   (: (+ ?*) (not (any ?/))))) (+ ?*) ?/)
+                    (: "," (group (* (any blank ?\C-j)) (any ?\} ?\] )))))
```
nbfalcon, line 44:
```suggestion
                    (: "," (group (* (any blank ?\C-j)) (any ?\} ?\])))))
```
remove space

nbfalcon (commented):
Looks good, but it will break all test cases, since they use dap-launch-remove-comments. Could you also fix those and add one for that case? To run the suite, use cask exec ert-runner.

jrwats:
Hang tight, trying to run these tests locally... back to "draft"

nbfalcon:
In `dap-launch-test--delete-commas', you should use let*.

dap-launch.el
```diff
@@ -28,8 +28,8 @@
 (require 'f)
 (require 'rx)
 
-(defun dap-launch-remove-comments ()
-  "Remove all C-style comments in the current buffer.
+(defun dap-launch-sanitize-json()
```
nbfalcon, line 31: Add a space before ().

test/dap-test.el
```diff
@@ -139,21 +139,34 @@
...
+         (post-exp (dap-launch-test--sanitize-json orig)))
     (should (string= post-exp "
 
 
 
 \"// string\"
 \"/*string*/\""))))
 
+(ert-deftest dap-launch-test--delete-commas ()
+  (let ((orig "{
```
nbfalcon, line 151:
```suggestion
  (let* ((orig "{
```
you're using earlier variables in let, meaning you must use let*.

jrwats: 🤦

nbfalcon:
Thank you for contributing to dap-mode and congratulations on your first PR!

jrwats:
> Thank you for contributing to dap-mode and congratulations on your first PR!

Thanks for the review!

dap-launch.el
```diff
@@ -28,8 +28,8 @@
 (require 'f)
 (require 'rx)
 
-(defun dap-launch-remove-comments ()
-  "Remove all C-style comments in the current buffer.
+(defun dap-launch-sanitize-json ()
+  "Remove all C-style comments and trailing commas in the current buffer.
```
kiennq, line 32: Shouldn't you need to go to `(point-min)` before doing `re-search-forward`?
Also wrap this one inside `save-excursion` too?

dap-launch.el
```diff
@@ -40,12 +40,13 @@ supported."
            (or (group
                 (or (: "//" (* nonl) eol)
                     (: "/*" (* (or (not (any ?*))
-                                   (: (+ ?*) (not (any ?/))))) (+ ?*) ?/)))
+                                   (: (+ ?*) (not (any ?/))))) (+ ?*) ?/)
+                    (: "," (group (* (any blank ?\C-j)) (any ?\} ?\])))))
```
kiennq, line 44: why not just use `space` instead of `blank and ?\n` only? In Windows, the newline characters can also include `\r`, sometimes even vertical space `\v` existed too. Wouldn't it safer to just check for all `space` group instead?

dap-launch.el
nbfalcon: From the docstring:
Replacement starts at point, and strings before it are ignored,
so you may want to move point to `point-min' with `goto-char'
first. This function moves `point'. Both // and /**/ comments are
supported.

That is as intended.

dap-launch.el
nbfalcon: Actually, this might be an issue, given that this isn't just JSON that we are parsing with its limited subset of whitespace, but JSON5, which supports more.

cc @jrwats

jrwats: Agreed. That `:blank:` misses other characters was an oversight on my part.  `:space:` won't match newline, however, that will still need `?\n` alongside it.

jrwats: https://github.com/emacs-lsp/dap-mode/pull/383

I'm really scratching my head about the `[[:space:]]*` vs `[[:space:]\n]*` repro...

PR #382, opened 2020-10-10
Bugfix/dap mouse broken#367

nbfalcon:
Make dap's hover tooltips not broken. Please see the REVIEW comments also.

### DIFF
--- a/dap-mouse.el
+++ b/dap-mouse.el
@@ -197,34 +197,35 @@ This function must return nil if it doesn't handle EVENT."
                             (list :expression expression
                                   :frameId active-frame-id
                                   :context "hover"))
-         (dap--resp-handler
-          (-lambda ((&hash "body" (var &as &hash? "result")))
-            (when (and (= request-id dap-tooltip--request)
-                       ;; REVIEW apparently ptsvd sends empty results?
-                       (not (string-empty-p result)))
-              (add-text-properties
-               start end '(mouse-face dap-mouse-eval-thing-face))
-              (when (get-buffer dap-mouse-buffer)
-                (kill-buffer dap-mouse-buffer))
-              (-let [(&hash "result" "variablesReference" variables-reference) var]
-                (if (= variables-reference 0)
-                    (message result)    ; primitive value -> posframe awkward
-                  ;; REVIEW I inherited this code; why kill the buffer first,
-                  ;; then show it, and then initialize it?
-                  (apply #'posframe-show dap-mouse-buffer
-                         :position start
-                         :accept-focus t
-                         dap-mouse-posframe-properties)
-                  (with-current-buffer (get-buffer-create dap-mouse-buffer)
-                    (lsp-treemacs-render
-                     `((:key ,expression
-                        :label ,result
-                        :icon dap-field
-                        :children ,(-partial #'dap-ui-render-variables
-                                             debug-session
-                                             variables-reference)))
-                     "" nil (buffer-name)))))
-              (add-hook 'post-command-hook 'dap-tooltip-post-tooltip))))
+         (-lambda ((&hash "success" "body" (var &as &hash? "result")))
+           (when (and success
+                      ;; REVIEW: hover failure will yield weird errors involving
+                      ;; process filters, so I resorted to this hack; we should
+                      ;; perhaps do proper error handling?
+                      (= request-id dap-tooltip--request))
+             (add-text-properties
+              start end '(mouse-face dap-mouse-eval-thing-face))
+             ;; REVIEW: I inherited this code; why kill the buffer first,
+             ;; then show it, and then initialize it?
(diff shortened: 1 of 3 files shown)


dap-mouse.el
```diff
@@ -197,32 +197,35 @@ This function must return nil if it doesn't handle EVENT."
...
+                      ;; perhaps do proper error handling?
+                      (= request-id dap-tooltip--request))
+             (add-text-properties
+              start end '(mouse-face dap-mouse-eval-thing-face))
+             ;; REVIEW: I inherited this code; why kill the buffer first,
+             ;; then show it, and then initialize it?
+             (when (get-buffer dap-mouse-buffer)
+               (kill-buffer dap-mouse-buffer))
+             (-let [(&hash "result" "variablesReference" variables-reference) var]
+               (if (= variables-reference 0)
```
yyoncho, line 213: use zerop

dap-mouse.el
```diff
@@ -197,32 +197,35 @@ This function must return nil if it doesn't handle EVENT."
...
+                   (message result)     ; primitive value -> posframe awkward
```
yyoncho, line 214: why is that? If it is primitive it still can be displayed in the posframe.

dap-mouse.el
nbfalcon: That line is gone, and I used zerop everywhere else.

dap-mouse.el
nbfalcon: Fixed it.

nbfalcon:
There is one more issue: for some reason, after hovering, the buffer gets marked as modified.

dap-mouse.el
```diff
@@ -197,32 +197,36 @@ This function must return nil if it doesn't handle EVENT."
...
-               (message message))))
+         (-lambda ((&hash "success" "body" (var &as &hash? "result")))
+           (when (and success
+                      ;; REVIEW: hover failure will yield weird errors involving
+                      ;; process filters, so I resorted to this hack; we should
+                      ;; perhaps do proper error handling?
+                      (= request-id dap-tooltip--request))
+             (add-text-properties
+              start end '(mouse-face dap-mouse-eval-thing-face))
+             ;; REVIEW: I inherited this code; why kill the buffer first,
```
yyoncho, line 208: Without this, in some cases, the posframe was not calculating the proper size of the tooltip.

yyoncho (approved):

dap-mouse.el
nbfalcon: Ahh, that makes sense. Thank you.

nbfalcon:
Dont merge yet.

nbfalcon:
How can I debug json messages between the server and Emacs? For some reason, when hovering over a variable that isn't defined yet, ptvsd sends a success response + empty result. This would cause empty hover boxes, so I check if the result isn't empty. This way, no empty box is shown, but the identifier is still highlighted. Should I change that?

nbfalcon:
@yyoncho, what do you think? Should I keep this workaround, should I perhaps also not highlight the identifier?

yyoncho:
> @yyoncho, what do you think? Should I keep this workaround, should I perhaps also not highlight the identifier?

I dont see this as a big issue. 

> How can I debug json messages between the server and Emacs? 

Set dap-print-io to t

nbfalcon:
@yyoncho I have it set to t, where would I find its result buffer? (Not restarted Emacs). Also, the last issue: why does the debugged buffer get marked as modified when I hover over a variable?

kiennq:
The log should be spewed on `*Message*`

nbfalcon:
Thanks, I can now conclude that I was right: ptsvd is weird. Here is the response for a hover on a yet-to-be-defined variable:
```json
{
  "type": "response",
  "seq": 21,
  "request_seq": 15,
  "success": true,
  "command": "evaluate",
  "message": "",
  "body": {
    "result": "",
    "variablesReference": 0,
    "presentationHint": {}
  }
}
```
"success": true, "variablesReference" = 0 -> no children and "result" is the empty string...

nbfalcon:
All that is missing now is to determine why the debugged buffer is modified when showing the posframe. Any ideas, anyone? Perhaps it is related to doom?

nbfalcon:
Im at a loss for my last point. Perhaps @kieenq might have some more ideas on how to not mark the buffer as modified.

kiennq:
> Im at a loss for my last point. Perhaps @kieenq might have some more ideas on how to not mark the buffer as modified.

There's `(set-buffer-modified-p FLAG)` if you want to mark the buffer as not modified. Can you try with vanilla Emacs to see if it's problem of `doom` or not?

nbfalcon:
I give up; slightly broken mouse support is still better than completely broken mouse support, so I'll merge it anyway.

PR #384, opened 2020-10-11
dap-python: Support attach mode

f279801:
- New configuration attribute `processId` is added.
  This name is picked to provide a consistent interface for both debugpy
  and ptvsd as well as make it works with launch.json naturally.

### DIFF
--- a/dap-python.el
+++ b/dap-python.el
@@ -194,7 +194,8 @@ overriden in individual `dap-python' launch configurations."
                       (plist-get conf :program)
                       (buffer-file-name)))
          (module (plist-get conf :module))
-         (debugger (plist-get conf :debugger)))
+         (debugger (plist-get conf :debugger))
+         (targetPid (plist-get conf :processId)))
     (cl-remf conf :debugger)
     (pcase (or debugger dap-python-debugger)
       ('ptvsd
@@ -205,14 +206,17 @@ overriden in individual `dap-python' launch configurations."
          (when (sequencep python-args)
            (setq python-args (mapconcat #'shell-quote-argument python-args " ")))
          (plist-put conf :program-to-start
-                    (format "%s%s -m ptvsd --wait --host %s --port %s%s %s %s"
+                    (format "%s%s -m ptvsd --wait --host %s --port %s %s"
                             (or dap-python-terminal "")
                             (shell-quote-argument python-executable)
                             host
                             debug-port
-                            (if module (concat " -m " (shell-quote-argument module)) "")
-                            (shell-quote-argument program)
-                            python-args))
+                            (if targetPid
+                                (format "--pid %s" targetPid)
+                              (format "%s %s %s"
+                                      (if module (concat " -m " (shell-quote-argument module)) "")
+                                      (shell-quote-argument program)
+                                      python-args))))
          (plist-put conf :debugServer debug-port)
          (plist-put conf :port debug-port)
          (plist-put conf :hostName host)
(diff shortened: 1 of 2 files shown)


docs/page/configuration.md
```diff
@@ -109,6 +109,31 @@ settings.
...
+            :request "attach"
+            :processId "${command:pickProcess}"
+            :name "Python :: Attach to running process"))
+    ```
+    The `${command:pickProcess}` configuration variable used by default to facilitate the user
+    selecting the debug process by a completion popping up window. The real `processId` can
+    always be specified using its *pid*.
+
+    **Note**: on Linux this is achieved using the [ptrace syscall](https://www.man7.org/linux/man-pages/man2/ptrace.2.html "ptrace syscall man page")
+    wrapped inside the gdb tool. Which mean below additional steps are also need to be done as well:
```
nbfalcon, line 126:
```suggestion
    wrapped inside the GDB tool, which means that for distributions that enable YAMA (e.g. Ubuntu) some additional steps may be necessary:
```

On this paragraph: shouldn't we state this is specific to distributions that use YAMA, like Ubuntu? From this thread I could gather that Fedora doesn't enable it by default: https://www.reddit.com/r/Fedora/comments/aeu3rj/yama_lsm_ptrace_restriction_disabled_by_default/.

dap-python.el
```diff
@@ -205,14 +206,17 @@ overriden in individual `dap-python' launch configurations."
          (when (sequencep python-args)
            (setq python-args (mapconcat #'shell-quote-argument python-args " ")))
          (plist-put conf :program-to-start
-                    (format "%s%s -m ptvsd --wait --host %s --port %s%s %s %s"
+                    (format "%s%s -m ptvsd --wait --host %s --port %s %s"
                             (or dap-python-terminal "")
```
nbfalcon, line 210:
```suggestion
                            (or (and dap-python-terminal (concat " " dap-python-terminal) ""))
```

nbfalcon (approved):
Looks good.

docs/page/configuration.md
```diff
@@ -109,6 +109,31 @@ settings.
...
+      - Install GDB.
+      - Turn on classic ptrace permission
+        ```bash
+        sudo sh -c 'echo 0 > /proc/sys/kernel/yama/ptrace_scope'
+        ```
+        Or starting the debugger under root user
+        ```elisp
+        setq dap-python-terminal "sudo "
```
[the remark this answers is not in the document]
f279801: sure, it should be.
Not sure what happen with my rebasing.

docs/page/configuration.md
f279801: agree, it's better to say it that way.

dap-python.el
f279801, line 210: Agree, but maybe swapping the space with `dap-python-terminal`
```suggestion
                            (or (and dap-python-terminal (concat dap-python-terminal " ") ""))
```

dap-python.el
```diff
@@ -205,14 +206,17 @@ overriden in individual `dap-python' launch configurations."
          (when (sequencep python-args)
            (setq python-args (mapconcat #'shell-quote-argument python-args " ")))
          (plist-put conf :program-to-start
-                    (format "%s%s -m ptvsd --wait --host %s --port %s%s %s %s"
+                    (format "%s%s -m ptvsd --wait --host %s --port %s %s"
```
yyoncho, line 209: @nbfalcon not part of this review, but do we support vterm here?

nbfalcon: @yyoncho we could use vterm here, but not with the facilities that I have already implemented for "runInTerminal", as they depend on a debug-session. However, adding vterm support to dap-debug shouldn't be that much effort (basically `let`-bind vterm-shell to :program-to-start and then call (vterm)). I could implement it if I have some more spare time.

nbfalcon: It should be max. 20 lines of Elisp, probably 4-10.

dvzubarev:
Any chance that this will be merged?

PR #386, opened 2020-10-13
dap-python: support debugging of 3rd modules

f279801:
- Add configuration handling to support step into 3rd modules.
- justMyCode is the single configuration needed to be specified
  regardless the selected back end debugger.
- DebugStdLib is also supported for backward compatibility, and
  a warning message will be printed suggesting for a migration
  if debugpy is selected as the debugger.


docs/page/configuration.md
```diff
@@ -105,6 +106,7 @@ settings.
             :cwd nil
             :env '(("DEBUG" . "1"))
             :target-module (expand-file-name "~/src/myapp/.env/bin/myapp")
+            :justMyCode :json-false
```
nbfalcon, line 109:
```suggestion
:justMyCode nil
```
`:justMyCode` should be optional and off by default. Also, use, and make sure the user can use, `nil` in such cases since that is more idiomatic in ELisp, which lacks a dedicated false type. json-encoder-specifics shouldn't leak trough.

dap-python.el
```diff
@@ -238,7 +245,12 @@ overriden in individual `dap-python' launch configurations."
          (cl-remf conf :module))
        (unless (plist-get conf :cwd)
          (cl-remf conf :cwd))
-
+       (when (and debug-options
+                  (sequencep debug-options)
+                  (seq-contains debug-options "DebugStdLib")
+                  (null justmycode))
+         (warn "DebugStdLib is deprecated, use justMyCode, please consider to update your configuration.")
```
nbfalcon, line 252: Why the backwards compatibility considerations here? `:debugOptions` is introduced with this PR, so there cannot be a config in the wild using it.

dap-python.el
```diff
@@ -216,7 +218,12 @@ overriden in individual `dap-python' launch configurations."
          (plist-put conf :debugServer debug-port)
          (plist-put conf :port debug-port)
          (plist-put conf :hostName host)
-         (plist-put conf :host host)))
+         (plist-put conf :host host)
+         (when (and (not (null justmycode))
```
nbfalcon, line 222:
```suggestion
         (when (and justmycode
```
Chaining `not` and `null` if only the result's truthynesss matters is redundant and non-idiomatic.

nbfalcon: FYI, `not` and `null` are aliases.

dap-python.el
nbfalcon: I'd just drop that warning altogether. You could state in the README somewhere that :debugOptions is `ptsvd`-specific, and that `:justMyCode` is the preferred way to go about this, since it is portable.

dap-python.el
```diff
@@ -216,7 +218,12 @@ overriden in individual `dap-python' launch configurations."
...
+                    (equal justmycode :json-false))
+             (setq debug-options (cl-pushnew "DebugStdLib" debug-options :test #'equal))
+             (plist-put conf :debugOptions debug-options)
```
nbfalcon, line 225: This will create a duplicate if :debugOptions is specified, with this version being overriden (it will show up twice in the JSON communcation). '(:debugOptions ...) -> '(:debugOptions (DebugStdLib . ...) :debugOptions ...)

nbfalcon:
My general suggestion here is: for `debugpy`, nothing else needs to be done (i.e.: no depreciation warnings, ...). :debugOptions should just be ignored, and not even removed. :justMyCode is alreay handled, implicitly for `debugpy`. For `ptsvd`, you can add additional support for `:justMyCode`. For this, you'll need to `plist-get` it, and to `cl-remf` it if `justMyCode` is explicitly specified.

PR #395, opened 2020-11-01
Fix `dap-variables-find-vscode-config`

jrwats:
I recently hit this issue when trying to launch `dap-debug`.  Attempting to read from `.vscode` hit errors when trying to read the value `t`.  Instead of returning `t` when a file exists, return the actual file.

https://github.com/emacs-lsp/dap-mode/commit/9ad62ea562f3651bda3f6adb94d3274811147ec9#r43781039

### DIFF
--- a/dap-variables.el
+++ b/dap-variables.el
@@ -506,7 +506,7 @@ for F in either ROOT/ or ROOT/.vscode/."
   (let* ((root (file-name-as-directory root))
          (root-f (concat root f))
          (root-vscode (concat root (file-name-as-directory ".vscode") f)))
-    (cl-some #'file-exists-p (list root-f root-vscode))))
+    (seq-find #'file-exists-p (list root-f root-vscode))))
 
 (defconst dap-variables-os-property-alist
   '((windows-nt . :windows)


jrwats:
@nbfalcon - fix here.  I still need to squash these stale commits, however...

nbfalcon:
Thank you, what a blunder on my part.

dap-variables.el
```diff
@@ -506,7 +506,7 @@ for F in either ROOT/ or ROOT/.vscode/."
   (let* ((root (file-name-as-directory root))
          (root-f (concat root f))
          (root-vscode (concat root (file-name-as-directory ".vscode") f)))
-    (cl-some #'file-exists-p (list root-f root-vscode))))
+    (seq-find #'file-exists-p (list root-f root-vscode))))
```
nbfalcon, line 509: Why seq-find over cl-find-if?

yyoncho: As a side note, in lsp-mode/dap-mode we favor dash and seq when it is not clear if we will get list or vector.

nbfalcon: @yyoncho that makes sense. However, in this case, we know its a list (because of the list constructor above).

nbfalcon: seq-find seems to even be a bit faster, so its better here.

jrwats: Yeah so `seq-*` helpers also have the benefit of being built-in to Emacs as of 25.1:

https://github.com/emacs-lsp/dap-mode/commit/6d57f07850d1960e48534b639cc718a1cfc8a409#commitcomment-43804821

jrwats: > Why seq-find over cl-find-if?

The truth is I actually couldn't find `cl-find-if` and Googling found me `seq-find` FWIW

yyoncho: > @yyoncho that makes sense. However, in this case, we know its a list (because of the list constructor above).

seq is the second choice, cl-* is not on the table at all. The usage of cl- is accidental.

nbfalcon: We should use `seq` where there are user-controlled values; however, in this case, we know its a list, and cl-find-if is slightly faster.

yyoncho: @nbfalcon if the performance is an issue, this code should not use the collection function at all, right? `cl-find-if` should not be used anywhere in favour of dash/seq equivalent to keep the code consistent.

Issue #402, opened 2020-11-12
Using `exwm`, debugging a floating X window GUI program causes display failure.


WJCFerguson:
`exwm`, the Emacs X Window Manager, will display a floating window in its own frame, so if the program you're debugging launches one, hits a breakpoint while that window has focus (typical), `dap-mode` creates its windows in _that_ frame, rather than where I'd expect (presumably the launched-from frame?).

I can think of a hack - I'd just make an advice function around `dap-debug` that notes the launched-from frame, then another advice around `dap-ui--show-buffer` to use that frame (`with-selected-frame`, I think).  But that's pretty grubby and not generic.  I wonder how it should be done more correctly?

I assume few people hit this combination.  If you have pointers though for how it *should* work, I might be able to generate a PR.

yyoncho:
I think that your solution is in the right direction, the issue that we should fix is to make it a bit more convenient(e. g. no advice but hooks). I would propose introducing dap-debug-started-hook + making dap-ui--show-buffer configurable so you could set it to your function. Other than that I am not sure what options do we have.

nbfalcon:
I propose something different - not showing `many-windows` in the stopped hook every time, but only once. I think I can implement this tomorrow.

PR #403, opened 2020-11-12
Add ui-display-frame to dap--debug-session struct

WJCFerguson:
Regarding issue #402.

Prior to this, when the debugger stops (e.g. breakpoint), it
spawns the ui elements in whatever is `(current-frame)`.  This changes
so that the UI display frame is set when the user starts or switches
to a session.

In particular the prior functionality is a bug for `exwm` users
debugging a program that produces a floating graphical window, which
will be contained in an Emacs frame under `exwm`.  Often a breakpoint
would be triggered by user interaction so that frame has focus.

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -264,6 +264,7 @@ The hook will be called with the session file and the new set of breakpoint loca
   (workspace nil)
   (threads nil)
   (thread-states (make-hash-table :test 'eql) :read-only t)
+  (ui-display-frame nil)
   (active-frame-id nil)
   (active-frame nil)
   (cursor-marker nil)
@@ -1457,7 +1458,8 @@ DEBUG-SESSIONS - list of the currently active sessions."
       (--each (dap--buffer-list) (with-current-buffer it
                                    (->> breakpoints
                                         (gethash buffer-file-name)
-                                        (dap--set-breakpoints-in-file buffer-file-name))))))
+                                        (dap--set-breakpoints-in-file buffer-file-name)))))
+    (setf (dap--debug-session-ui-display-frame new-session) (selected-frame)))
 
   (run-hook-with-args 'dap-session-changed-hook lsp--cur-workspace)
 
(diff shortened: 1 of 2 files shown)


dap-mode.el
```diff
@@ -264,6 +264,7 @@ The hook will be called with the session file and the new set of breakpoint loca
   (workspace nil)
   (threads nil)
   (thread-states (make-hash-table :test 'eql) :read-only t)
+  (ui-display-frame nil)
```
WJCFerguson, line 267: Obviously the first question is whether this is an appropriate place to store this info.  Second issue is naming?

kiennq: Do we want this so each session can has different display frame? Or if there's only one session, just a global `defvar` is enough?

WJCFerguson: If `dap-mode` can't accommodate multiple simultaneous sessions(?), and isn't ever expected to, then a `defvar` definitely seems more appropriate.  Otherwise I would expect you'd give each its own frame.

I'd be happy to make that change.  I'm just getting started with `dap-mode` as of yesterday and hit this issue.  Let me know, thanks!

nbfalcon: @WJCFerguson I'm assuming the windows causing issues are the ones created by `dap-ui-many-windows-mode` (which is enabled by default). In that case, your approach may be inappropriate: the `dap-ui-many-windows-mode` windows would be created in the same as the debug session, but others spawned manually would be too. So if someone starts a debug session, switches to a different frame and then does `M-x dap-ui-locals`, the locals window would be displayed in the old frame, not the new one, as one would probably expect.

As to your questions: `dap-mode` supports multiple parallel (running) debug sessions, which can be inspected using `dap-ui-sessions`.

Thanks for contributing to `dap-mode`!

WJCFerguson:
This was put up as much as anything to spur discussion.  I think it is not a suitable fix in the end so it may as well be closed.

PR #414, opened 2020-11-26
`dap-start-debugging-noexpand': expand :cwd

### LINKED ISSUE #413: dap-python, how to set cwd for debugpy when debugging a script
As title! I tried to set `:cwd` arg in my template but it gets ignored.

nbfalcon:
In a previous commit, :cwd was expanded specifically in `dap-chrome`. Undo that,
and centralize expanding :cwd in `dap-start-debugging-noexpand`. This allows
:cwd to be a relative path.

Fixes #413.

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -1590,7 +1590,8 @@ before starting the debug process."
                    :wait-for-port :type :request :port
                    :startup-function :environment-variables :hostName host) launch-args)
           (session-name (dap--calculate-unique-name name (dap--get-sessions)))
-          (default-directory (or cwd default-directory))
+          (default-directory (or (and cwd (expand-file-name cwd))
+                                 default-directory))
           (process-environment (if environment-variables
                                    (cl-copy-list process-environment)
                                  process-environment))
--- a/dap-chrome.el
+++ b/dap-chrome.el
@@ -50,7 +50,7 @@
   (setq conf (-> conf
                  (plist-put :type "chrome")
                  (plist-put :dap-server-path dap-chrome-debug-program)
-                 (dap--put-if-absent :cwd (expand-file-name default-directory))))
+                 (dap--put-if-absent :cwd default-directory)))
   (dap--plist-delete
    (pcase (plist-get conf :mode)
      ("url" (-> conf


dap-mode.el
```diff
@@ -1590,7 +1590,8 @@ before starting the debug process."
                    :wait-for-port :type :request :port
                    :startup-function :environment-variables :hostName host) launch-args)
           (session-name (dap--calculate-unique-name name (dap--get-sessions)))
-          (default-directory (or cwd default-directory))
+          (default-directory (or (and cwd (expand-file-name cwd))
```
yyoncho, line 1593: default-directory should be expanded too.

nbfalcon: Why that? From its doc: "it should be an absolute directory name". Also, against what would `default-directory` be expanded, given that we omit it here? The doc of `expand-file-name` says that its argument is expanded against `default-directory` unless the second argument is specified.

All this would achieve is that "." and ".." paths get eliminated. This way, we don't change `default-directory` at all unless `:cwd` is specified.

nbfalcon: Your fix for `dap-chrome` would still behave exactly the same way, because it specifies a `default-directory` `:cwd` that is then needlessly expanded against `default-directory`.

yyoncho: For the same reason we expand default-directory in dap-chrome.el - default directory could be `~/foo/bar`

yyoncho: Also, the PR is introducing regression - we will send `~/foo/bar` to the adapter since :cwd won't be expanded.

yyoncho: To elaborate a bit more, the proper fix is to expand default-directory as described in the root comment and then set cwd in the template.

nbfalcon: This will still work though:
```emacs-lisp
(let ((default-directory "~/Downloads")) (shell-command "pwd")) ;; => `message`s "/home/nikita/Downloads" on my system"
```

I think there are five options here:
- `expand-file-name` in `dap-chrome`
- always send an expanded :cwd, even if unspecified
- send an expanded :cwd only if specified
- Either of the above, but also expand `default-directory` (against itself)
:cwd would be expanded in all cases. I ended up going with the last one, always sending an expanded :cwd and always expanding `default-directory`.

yyoncho: The cwd should be set in the template otherwise commands like restart session won't work in the expected manner when cwd is relative.

nbfalcon: Why does `dap-start-debugging-noexpand` mutate LAUNCH-ARGS with `plist-put :name`?

yyoncho: What is the problem caused by that?

yyoncho: > Why does `dap-start-debugging-noexpand` mutate LAUNCH-ARGS with `plist-put :name`?

Modifying the template should be fine and expected since we are copying it.

PR #417, opened 2020-12-09
Added rust specific example setting

mslot:
I have added an update for rust. Rust strings is not built up around end escaping strings with \0. There for a native debugger like GDB will not show strings correctly in the Locals area when stepping through a debugging session. Rust has a tool for this which is basically some python formatters wrapped around gdb. All this is shipped with the rust toolchain when installing rust, so rust-gdb is on the PATH and ready to be called.

This brings me to my update of the configuration: add a gdbpath that tells the debugger to use rust-gdb instead of the plain version of gdb.

I have added this configuration so others can enjoy using dap mode while coding rust.


yyoncho:
Thank you!

PR #429, opened 2021-01-27
Specify window to resize rather than relying on the current window

genegoykhman:
When the `no-select` option is passed to `dap-go-to-output-buffer` the default window isn't changed (it is not set to the window containing the output buffer). This PR explicitly resizes the window with the output buffer even when the default window remains set to something else.


jcs090218 (approved):

jcs090218:
Good catch! Thanks! :)

PR #432, opened 2021-02-03
Added the ability to update debugger and remove the oldest

zakariaz:
Hi,
I implemented this function to allow the dynamic generation of specific dap update functions for different mode.
It checks if there is a debugger installed and compares both versions (mentioned in lsp and installed) then downloads the lsp version if is different from the previous one. Of course it will remove the installed one.
This functionality is aimed to provide the compatible version of the debugger Not the latest.

I would be happy to get feed back from you.

### DIFF
--- a/dap-utils.el
+++ b/dap-utils.el
@@ -142,7 +142,7 @@ With prefix, FORCED to redownload the extension." extension-name)))
 PUBLISHER is the vscode extension publisher.
 NAME is the vscode extension name.
 VERSION is the version to update to
-PATH is the location of directory that contains the debugger. It shoul end with the version number"
+PATH is the download destination dir.
   (let* ((extension-name (concat publisher "." name))
 	 (constructed-path (f-join dap-utils-extension-path
 			   "vscode"
--- a/dap-utils.el
+++ b/dap-utils.el
@@ -137,7 +137,7 @@ With prefix, FORCED to redownload the extension." extension-name)))
          (message "%s: %s debug extension are not set. You can download it with M-x %s-setup"
                   ,dapfile ,extension-name ,dapfile)))))
 
-(defmacro dap-utils-vscode-update-function (dapfile publisher name version &optional path callback)
+(defmacro dap-utils-vscode-update-function (dapfile publisher name version &optional path)
   "Helper to create DAPFILE update function for vscode debug extension.
 PUBLISHER is the vscode extension publisher.
 NAME is the vscode extension name.
--- a/dap-utils.el
+++ b/dap-utils.el
@@ -142,7 +142,7 @@ With prefix, FORCED to redownload the extension." extension-name)))
 PUBLISHER is the vscode extension publisher.
 NAME is the vscode extension name.
 VERSION is the version to update to
-PATH is the download destination dir.
+PATH is the download destination dir."
   (let* ((extension-name (concat publisher "." name))
 	 (constructed-path (f-join dap-utils-extension-path
 			   "vscode"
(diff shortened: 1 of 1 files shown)


yyoncho:
Thank you for the PR! 

At this point, there are two paths:

1. Work on this implementation towards merging it - the user will have to explicitly call update(I have few suggestions to improving way this command is exposed to the user).
2. Work on fixing dap-mode to do that automatically - in this case, when you run a particular debug template, dap-mode will check if the adapter is present/up-to-date and it will ask to install/update the server.

2) is something that is planned for a long time and it will ultimately improve UE. This is similar to lsp-mode asking the user to install the server instead of the user having to manually call `lsp-install-server`. Let me know how do you want to proceed.

zakariaz:
@yyoncho The second option is obviously a better solution. dap-mode should handle that automatically.
Any advice would be very helpfull.
I am also very interested in your suggestions to improve the submitted code. that will help me provide better work in the future.

Thank for reviewing my work

yyoncho:
> @yyoncho The second option is obviously a better solution. dap-mode should handle that automatically.
> Any advice would be very helpfull.

I am in favor of simple designs:

1. Introduce function which will be future replacement to `dap-register-debug-provider`. It will have:

```
(dap-register-debug-provider
 :language-id "FOO"
 :process-template-function (lambda (template)
                              :process-template-function (lambda (template)
                                                           ...))

 :present? (lambda ()
             ;; returns one of :upgrade :up-to-date :none
             )
 :install (lambda ()
            ;; performs install.
            ))
```

2. Then in here - https://github.com/emacs-lsp/dap-mode/blob/master/dap-mode.el#L1664 we will check if there is `:present?` function for this adapter, ask the user whether they want to install/update, and perform the install via `:install` function.

3. Create functions that will generate install/present? functions for our most common sources(e. g. marketplace . 

This is pretty basic, I am probably missing something so if you have questions/concerns/suggestions you could ping me here/gitter/discord.

> am also very interested in your suggestions to improve the submitted code. that will help me provide better work in the future.

Nothing useful - I just wanted an update and install functions to be somehow merged.

PS: I think that at some point we might want to merge this with `lsp-dependency`. the more complicated part is that `lsp-dependency-ensure` is async but I guess that this can be adjusted later assuming that it will be behind functions created in 3.

zakariaz:
I think questions and suggestions will rise when I start working on it. 
Thanks for your comment

zakariaz:
Hi @yyoncho,
I worked on a first [draft](https://github.com/zakariaz/dap-mode/blob/de6fb41b09f93f492d62b1d8a8402a24ebce8984/dap-mode.el#L1496) of the replacement function of `dap-register-debug-provider`.

it just takes named parameters and just stores them. i think that's what what was discussed earlier in requirement number 1 and 2.

Please take a look, your feedback would be very appreciated.

Thank you for your time

yyoncho:
@zakariaz can you utilize cl-defun?

zakariaz:
@yyoncho 
I refactor using `cl-defun`. You can see the change by clicking the link in the last comment

yyoncho:
> @yyoncho
> I refactor it using `cl-defun`. You can see the change by clicking the link in the last comment

I don't see it. Can you open a draft PR?

zakariaz:
I made a pull request to master. I think it's better to work on it that way

zakariaz:
Hi @yyoncho, i worked on a version that is working. Actually it passed all the tests. here is the [link](https://github.com/emacs-lsp/dap-mode/pull/441)

yyoncho:
Good, I added one comment. I think that after addressing it you may continue with integrating all this with dap-mode.el.

yyoncho:
I will close this PR. Lets use the draft PR for future discussion.

PR #436, opened 2021-02-17
Add adapter for CodeLLDB

puuuuh:


yyoncho (approved):
Looks good, there are several formatting issues that I will fix after merging the PR. 

We need to figure out which adapter.

PR #441, opened 2021-02-21
add a future replacement of dap-debug-provider

zakariaz:
The function `dap-register-setup-debug-provider` register 3 functions according to the provided language ID:
-  A process template function
- A function that checks if the debug extension is installed or outdated. 
- A function that performs install or update of the debug extension according to the user decision.

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -1591,6 +1591,9 @@ before starting the debug process."
                    :startup-function :environment-variables :hostName host) launch-args)
           (session-name (dap--calculate-unique-name name (dap--get-sessions)))
           (default-directory (or cwd default-directory))
+          (process-environment (if environment-variables
+                                   (cl-copy-list process-environment)
+                                 process-environment))
           program-process)
     (mapc (-lambda ((env . value)) (setenv env value)) environment-variables)
     (plist-put launch-args :name session-name)
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -1705,7 +1705,7 @@ after selecting configuration template."
                                                   (user-error "%s does not specify :type" debug-args))))))
     (cond
      ((string= provider-state :upgrade)    (funcall install provider-state debug-provider-name))  ;upgrade debug-provider
-     ((string= provider-state :none)       (functall install provider-state debug-provider-name)) ;install debug-provider
+     ((string= provider-state :none)       (funcall install provider-state debug-provider-name)) ;install debug-provider
      ((string= provider-state :up-to-date) nil)                                                   ;do nothing
      ((null provider-state)                (error (concat "None of :upgrade, :none, "
 					                  ":up-to-date was provided by "
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -18,7 +18,7 @@
 ;; Author: Ivan Yonchovski <yyoncho@gmail.com>
 ;; Keywords: languages, debug
 ;; URL: https://github.com/emacs-lsp/dap-mode
-;; Package-Requires: ((emacs "26.1") (dash "2.18.0") (lsp-mode "6.0") (bui "1.1.0") (f "0.20.0") (s "1.12.0") (lsp-treemacs "0.1") (posframe "0.7.0"))
+;; Package-Requires: ((emacs "26.1") (dash "2.18.0") (lsp-mode "6.0") (bui "1.1.0") (f "0.20.0") (s "1.12.0") (lsp-treemacs "0.1") (posframe "0.7.0") (ht "2.3"))
 ;; Version: 0.6
 
 ;;; Commentary:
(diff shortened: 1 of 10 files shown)


zakariaz:
Those two function are missing defined `dap-register-debug-provider`, `dap-register-debug-template`.
I have `dap-debug-provider` and `dap-debug-template` even if i'am tracking the upstream repo. the earlier functions doesn't appear in my fork!

I don't know what to do. I searched on the internet did a `git remote add upstream https://github.com/emacs-lsp/dap-mode.git`
and did `git fetch upstream`, `git merge upstream/master` and nothing seems to work

zakariaz:
> ```
> (defun dep-register-setup-debug-provider (&rest debug-setup)
> ```
> 
> debbug -> `debug` (debbug-setup will have to be renamed in the body as well).

i'v done it

dap-mode.el
```diff
@@ -1493,6 +1493,39 @@ has language id = LANGUAGE-ID. The function must return debug
...
+					      "Please refer to it documentation")))
+   ((null language-id)         (error "'langauge-id' is not provided"))
+   ((null process-template-fn) (error "'process-template-fn' is not provided"))
+   ((null present?)            (error "'present?' is not provided"))
+   ((null install)             (error "'install' is not provided"))
+   (t nil))
+  
+  (let ((alist-debug-setup (seq-partition debug-setup 2)))
+    (dolist (entry alist-debug-setup)
+      (puthash (car entry) (cadr entry) dap--debug-providers))))
```
yyoncho, line 1527: I think we should either use language-id as a key and associate it to plist/alist with the other keys or use composite key something like `(cons language-id :present?)` because when we register 2 providers we will have 2 entries in that alist and there won't be a way to distinguish what is the corresponding :present? function, right?

zakariaz: I would go for association of #( ... :language-id plist) it make more sense, we just ask for the key in the plist and we don't expect duplication of keys and values. for the composite key, i don't understand your suggestion. it would be something like this ((language-id :present) . present?) ?.

zakariaz:
@yyoncho i finished the implementation of the function it's something simple. Nothing fancy. But it works.

yyoncho:
> @yyoncho i finished the implementation of the function it's something simple. Nothing fancy. But it works.

Looks good.

zakariaz:
Hi @yyoncho,
I started to play with the function `dap-debug` to figure out how to refactor it to process the new `dap--debug-providers` hash table.

There some drastic changes that has to be done to the function,when using the `:type` retrieved from `debug-args` to get the  3 functions from `dap--debug-providers` there will be a plist and not a function anymore so there will be some breaking changes.
Do you have any advice before i start to refactor the function?
Also i don't know what you think about the usage `dap-register-setup-debug-provider`. I guess it will be used inside every specific dap mode (dap-node, dap-php ...) to be populated with the proper setup. 
Do you have another strategy to handle that case or should it be used that way?

zakariaz:
The `dap-utils-extension-present?` doesn't handle edge cases like when there is more than one extension in the directory and the extension version that we are checking for is present in that folder. it basically checks the latest version installed against the version provided by the dev.
I pushed it so someone can take a look and give feedback. I will improve it soon

zakariaz:
@yyoncho I've finished the required changes to dap-mode.
Waiting for review when you get time.

yyoncho:
Thank you, I will try to take a look tomorrow or during the weekend.

zakariaz:
You are welcome

dap-utils.el
```diff
@@ -85,6 +106,57 @@ PATH is the download destination path."
...
+    (lambda (provider-state debug-provider-name)
+      "Install or upgrade DEBUG-PROVIDER-NAME according to PROVIDER-STATE.
+Argument PROVIDER-STATE is either `:up-to-date' which instruct the function
+ to do nothing, `:upgrade' which instruct the function to upgrade or `:none'
+which tell the function to install the missing provider.
+Argument DEBUG-PROVIDER-NAME is the name of the language related to dap provider"
+      (cond
+       ((eq provider-state :up-to-date) (message "%s-debug-provider is already installed and up to date"
+						 debug-provider-name))
+       ((eq provider-state :upgrade)    (progn
```
yyoncho, line 148: nit: there is no need from progn.

yyoncho (commented):
Looks good, can you fix one of the debug adapters(e. g. JS one) so I can see how it is going to be used in practice?

zakariaz:
> Looks good, can you fix one of the debug adapters(e. g. JS one) so I can see how it is going to be used in practice?

sure. I will do it

dap-utils.el
zakariaz: i corrected it

zakariaz:
> Looks good, can you fix one of the debug adapters(e. g. JS one) so I can see how it is going to be used in practice?

sure. I will do it

PR #444, opened 2021-02-23
`dap-go': bump vscode-go version

### LINKED ISSUE #399: Cannot see go map in local variables
When debugging golang app and expanding local map variable, the following message is shown:

```
Wrong type argument: hash-table-p, nil
```

Map is not expanded. Viewing lists and channels work ok.

nbfalcon:
Version 0.15.0 of vscode-go does not handle maps correctly, adding nulls to
them. This bug is fixed in version 0.22.1, so bump it in `dap-go-setup`.

The layout of the extension changed: the executable that needs to be run is no
longer goDebug.js, bug dist/debugAdapter.js. As such, users will need to
manually update `dap-go` using `C-u dap-go-setup`.

Fixes #399.

@yyoncho as noted in the second paragraph, this is a breaking change. This is unavoidable though, as currently `dap-utils` doesn't version vscode-downloads. How should this be handled?

### DIFF
--- a/dap-go.el
+++ b/dap-go.el
@@ -41,7 +41,8 @@
            (new (f-join dap-go-debug-path "extension/dist/debugAdapter.js")))
        (if (f-exists? old)
            (progn
-             (lsp--warn "Go debug adapter is outdated; some features will not work properly")
+             (lsp--warn "Go debug adapter is outdated; some features will not work properly (map debugging).\n\
+Update `dap-go' using `C-u M-x dap-go-setup'")
              old)
          new)))
   "The path to the go debugger."
--- a/dap-go.el
+++ b/dap-go.el
@@ -35,8 +35,15 @@
   :group 'dap-go
   :type 'string)
 
-(defcustom dap-go-debug-program `("node"
-                                  ,(f-join dap-go-debug-path "extension/dist/debugAdapter.js"))
+(defcustom dap-go-debug-program
+  `("node"
+    ,(let ((old (f-join dap-go-debug-path "extension/out/src/debugAdapter/goDebug.js"))
+           (new (f-join dap-go-debug-path "extension/dist/debugAdapter.js")))
+       (if (f-exists? old)
+           (progn
+             (lsp--warn "Go debug adapter is outdated; some features will not work properly")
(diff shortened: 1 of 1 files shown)


yyoncho:
will be solved by https://github.com/emacs-lsp/dap-mode/pull/432

yyoncho (approved):

nbfalcon:
@yyoncho do I merge now or later?

yyoncho:
> @yyoncho do I merge now or later?

Can you add warning in populate if you are able to find the extension on the old path? That should be good enough for current users.

yyoncho:
I don't know what else we can do. If the current version is not working breaking the compatibility is justified.

nbfalcon:
@yyoncho do you think we should have a warning if the old version is installed, or to silently handle it? It would just a be a simple f-exists check in `dap-go--populate-start-file-args`.

yyoncho:
> do you think we should have a warning if the old version is installed,

I think that will be helpful for users that have to migrate.

nbfalcon:
@yyoncho how should commands be written in `lsp--warn`? What do you think about "Update `dap-go' using `C-u M-x dap-go-setup`"? Other than that, the issue I've been having with debug sessions not starting went away, so this could be merged.

yyoncho:
> @yyoncho how should commands be written in lsp--warn? What do you think about "Update dap-go' using C-u M-x dap-go-setup`"? 

Souds good to me.

> Other than that, the issue I've been having with debug sessions not starting went away (I don't know what was happening), so this could be merged:

I don't see where the issue is - can you give me more pointers?

nbfalcon:
For some time, I could not successfully start go debug sessions. Not having changed anything, everything (Launch executable/file) works correctly now, so there is no issue.

nbfalcon:
Now I think I know the problem: with certain options (`--trace-depreciations --trace-warnings` (also can be set in `NODE_OPTIONS`)), nodejs will throw an error whenever a deprecated function is called (`new Buffer()` is called in the latest version of the adapter). I guess we'll need to switch to dlv eventually, but this is out of the scope of this PR.

PR #448, opened 2021-03-16
:compilation for `dap-debug`

nbfalcon:
`dap-debug` now respects a `:compilation` option, which is a pre-compile step as
a `compilation` shell command (as discussed on discord with @yyoncho). This is
more of a POC and probably needs more refinement:

- [x] There is a "debug session exited" message printed for reason some (I was using "compilation", not "dap-compilation")
- [x] Documentation
- [ ] (out of scope?) task runner integration

----

#

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -1686,7 +1686,8 @@ be used to compile the project, spin up docker, ...."
     (-if-let ((&plist :dap-compilation compilation) launch-args)
         (progn
           (cl-remf launch-args :dap-compilation)
-          (with-current-buffer (compilation-start compilation)
+          (with-current-buffer (compilation-start compilation t (lambda (&rest _)
+                                                                  "*DAP compilation*"))
             (let (window)
               (add-hook 'compilation-finish-functions
                         (lambda (buf status &rest _)
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -1689,7 +1689,8 @@ after selecting configuration template."
                           (if (string= "finished\n" status)
                               (progn
                                 (when (and (not dap-debug-compilation-keep)
-                                           (window-live-p window))
+                                           (window-live-p window)
+                                           (eq buf (window-buffer window)))
                                   (delete-window window))
                                 (funcall cb))
                             (lsp--error "Compilation step failed"))))
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -1646,7 +1646,7 @@ before starting the debug process."
         (run-hook-with-args 'dap-session-created-hook debug-session)))))
 
 (defcustom dap-debug-compilation-keep nil
-  "Whether `dap-debug' should keep the compile window on sucess.
+  "Whether `dap-debug' should keep the compile window on success.
 By default, it is hidden."
   :type 'boolean
   :group 'dap-mode)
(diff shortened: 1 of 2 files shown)


dap-mode.el
```diff
@@ -1668,10 +1674,27 @@ after selecting configuration template."
...
+          (cl-remf launch-args :compilation)
+          (with-current-buffer (compilation-start compilation)
+            (let (window)
+              (add-hook 'compilation-finish-functions
+                        (lambda (buf status &rest _)
+                          (if (string= "finished\n" status)
+                              (progn
+                                (when (and (not dap-debug-compilation-keep)
+                                           (window-live-p window))
+                                  (delete-window window))
```
yyoncho, line 1693: can we check if `get-buffer-window` in compilation buffer is non-nil and kill it in that way? Otherwise, if user displays a different buffer in that window we will still close it.

dap-mode.el
```diff
@@ -1668,10 +1674,27 @@ after selecting configuration template."
...
-        (funcall launch-args #'dap-start-debugging-noexpand)
-      (dap-start-debugging-noexpand launch-args))))
+                                          (user-error "%s does not specify :type" debug-args)))))
+         (cb (lambda ()
+               (if (functionp launch-args)
+                   (funcall launch-args #'dap-start-debugging-noexpand)
+                 (dap-start-debugging-noexpand launch-args)))))
+    (-if-let ((&plist :compilation) launch-args)
+        (progn
+          (cl-remf launch-args :compilation)
```
yyoncho, line 1684: nit: can you move it in `dap-start-debugging-noexpand` near the other cleanup? Also, I think that it will be better if we can use :dap-compilation to avoid collisions with valid keys template keys.

yyoncho (approved):
Looks good, few minor nits:

dap-mode.el
nbfalcon: Fixed.

dap-mode.el
nbfalcon: Are you sure it `dap-start-debugging-noexpand` is the right function for this? Isn't it supposed to be a more low-level interface?

nbfalcon: Second point is fixed.

yyoncho: it already deletes the other custom stuff like  `:cleanup-function` it seems like it is better to keep the cleanup in one place.

nbfalcon: The problem with using :cleanup-function is that it would have to have a closure over the compilation window (since we don't want to delete the compilation buffer if the user opens it manually), and that wouldn't work with `dap-debug-last`.

yyoncho: I am talking solely about removing :compilation key from the plist.

yyoncho: ... and the location of that code. 

> (cl-remf launch-args :compilation)

nbfalcon: Ahh, I understand now. However, `:dap-compilation` wouldn't be treated specially unlike the other fields when using `dap-start-debugging-noexpand`, so I think it makes sense to only remove it in `dap-debug`.

yyoncho: My understanding is  that we remove keys to avoid sending them to the debug adapter. ATM this is happening in `dap-start-debugging-noexpand`. Also, you should not remove it inplace from the configuration othewise it will be removed from the rerun.

nbfalcon: We wouldn't modify the configuration, because `dap-variables-expand-in-launch-configuration` necessarily copies the tree passed to it. I just realized that we are needlessly copying the conf in the `interactive` block.

yyoncho (approved):

PR #470, opened 2021-05-05
Fix custom type mismatches

wyuenho:
Symbols inside consts dont't need to be quoted and nil is not a valid value for the string type


jcs090218 (approved):
LGTM

yyoncho (approved):

yyoncho:
Thank you!

PR #472, opened 2021-05-13
dap-debug: remove compilation hook after use

kendatsuba:
Prevent hooks from piling up each time dap-debug is called with
'dap-compilation' set in launch.json. If they do, multiple debug
sessions will be started after a successful compilation, not necessarily
linked to the current debug target but to previous ones.


yyoncho (approved):
Nice catch!

jcs090218 (approved):

PR #477, opened 2021-06-04
Remove custom space

jcs090218:
Remove custom elpa path.

Same as PR, https://github.com/emacs-lsp/lsp-mode/pull/2908

### DIFF
--- a/test/windows-bootstrap.el
+++ b/test/windows-bootstrap.el
@@ -23,12 +23,6 @@
 
 (require 'package)
 
-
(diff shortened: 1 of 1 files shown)


kiennq:
Is there a reason why we want to not use the custom path? The custom path is there to allow to run the test locally without overwriting the personal Emacs setup.

jcs090218:
The CI fails for some reason when having a custom path. I am still trying to investigate it, yet I don't think having a custom path is necessary for a CI test? I would assume user will just use the default path from Emacs.

Issue #54, opened 2019-03-09
add functions for going up/down the stack


yyoncho:

prittjam:
has anyone addressed this, I was just about to suggest the same.

yyoncho:
No, not it is still open. If you want to give it a try you may take a look at `dap-switch-stack-frame` and/or ping me here or in gitter.

prittjam:
Unfortunately, me elisp is barely existent; otherwise I would give it a go.

PR #485, opened 2021-07-05
Add dap-up-stack-frame and dap-down-stack-frame

gvol:
This is to fix https://github.com/emacs-lsp/dap-mode/issues/54

1. I used `-if-let*` instead of a bunch of nested `-if-let`s.  I feel this made the code much easier to read, at the expense of better error messages.  I'm open to changing it if you think the better error messages are worth it.
2. I didn't add any tests.  I did briefly glance at the tests, but I didn't find one testing `dap--go-to-stack-frame` which I would have liked to copy.  Let me know if you have any pointers on how to write some good tests.
4. I have only tested this with `dap-lldb`.
3. I added it to the hydra under the "Switch" section rather than the "Stepping" section.  I kind of wanted to put it to "Stepping", so that it would be easier to invoke, but it seemed like it fit better for the switching (given that's where the generic stack frame switching is).

Please let me know your thoughts, and thank you so much for allowing me to finally debug inside Emacs on Mac OS!

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -1335,6 +1335,29 @@ RESULT to use for the callback."
              error))
     (error "No thread is currently active %s" (dap--debug-session-name (dap--cur-session)))))
 
+(defun dap-up-stack-frame (frames)
+  "Switch stackframe up FRAMES frames on the current thread.
+A negative value will move down frames."
+  (interactive "p")
+  (-if-let* ((session (dap--cur-session-or-die))
+             (thread-id (dap--debug-session-thread-id session))
+             (stack-frames (gethash thread-id
+                                    (dap--debug-session-thread-stack-frames session)))
+             (cur-frame (dap--debug-session-active-frame session))
+             (pos (cl-position cur-frame stack-frames))
+             (len (length stack-frames))
+             (new-pos (min (1- len) (max 0 (+ pos frames)))))
+      (if (eq pos new-pos)
+          (error "Already at the %s of the stack" (if (<= pos 0) "bottom" "top"))
+        (dap--go-to-stack-frame session (nth new-pos stack-frames)))
+    (error "Unable to find active session, thread, or frame.")))
+
+(defun dap-down-stack-frame (frames)
+    "Switch stackframe down FRAMES frames on the current thread.
+A negative value will move up frames."
+  (interactive "p")
+  (dap-up-stack-frame (- frames)))
+
 (defun dap--calculate-unique-name (debug-session-name debug-sessions)
   "Calculate unique name with prefix DEBUG-SESSION-NAME.
 DEBUG-SESSIONS - list of the currently active sessions."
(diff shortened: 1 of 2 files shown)


dap-mode.el
```diff
@@ -1335,6 +1335,29 @@ RESULT to use for the callback."
              error))
     (error "No thread is currently active %s" (dap--debug-session-name (dap--cur-session)))))
 
+(defun dap-up-stack-frame (frames)
+  "Switch stackframe up FRAMES frames on the current thread.
+A negative value will move down frames."
+  (interactive "p")
+  (-if-let* ((session (dap--cur-session))
```
yyoncho, line 1342: dap--current-session-or-die?

yyoncho (commented):
Thank you for the PR!



> 1. I used `-if-let*` instead of a bunch of nested `-if-let`s.  I feel this made the code much easier to read, at the expense of better error messages.  I'm open to changing it if you think the better error messages are worth it.

Generally, you could use the following construct:

```
(let ((a (or (b) 
                 (error "b returned null"))
    ....
    )
```

I am fine with the code as it is.

> 2. I didn't add any tests.  I did briefly glance at the tests, but I didn't find one testing `dap--go-to-stack-frame` which I would have liked to copy.  Let me know if you have any pointers on how to write some good tests.

Currently, the tests do not run (at least the integration one) mainly due to the instability of the JDTLS server but I haven't tried to make them work again for 2 years or more.

> 3. I have only tested this with `dap-lldb`.

That's fine.

> 4. I added it to the hydra under the "Switch" section rather than the "Stepping" section.  I kind of wanted to put it to "Stepping", so that it would be easier to invoke, but it seemed like it fit better for the switching (given that's where the generic stack frame switching is).
 
Can you post a screenshot how it looks like?

dap-mode.el
```diff
@@ -1335,6 +1335,29 @@ RESULT to use for the callback."
...
+             (thread-id (dap--debug-session-thread-id session))
+             (stack-frames (gethash thread-id
+                                    (dap--debug-session-thread-stack-frames session)))
+             (cur-frame (dap--debug-session-active-frame session))
+             (pos (position cur-frame stack-frames))
```
yyoncho, line 1347: CI breaks, please use cl-position.

dap-mode.el
gvol: Good call.

dap-mode.el
gvol: Oops. I guess it's time for me to remove the `(require 'cl)` from my .emacs.

gvol:
Here's a screen shot of the updated hydra: Screen Shot 2021-07-06 at 10 40 44

I took the liberty of force pushing the changes since they were small and that's how we usually operate where I work.  I hope that's not a problem.  If you'd prefer not, I'll try to remember in the future.

yyoncho (approved):

yyoncho:

PR #489, opened 2021-07-13
Add Erlang support

robertoaloi:


dap-erlang.el
```diff
@@ -0,0 +1,43 @@
...
+;;; Code:
+
+(require 'dap-mode)
+
+(defun dap-erlang--populate-start-file-args (conf)
+  "Populate CONF with the required arguments."
+  (-> conf
+    (plist-put :dap-server-path '("els_dap"))
+    (plist-put :projectDir (lsp-find-session-folder (lsp-session) (buffer-file-name)))
+    (plist-put :cwd (lsp-find-session-folder (lsp-session) (buffer-file-name)))
```
yyoncho, line 38: why not dap--put-if-absent for these two?

robertoaloi: The `cwd` one in VS Code often contains `${workspaceRoot}` which is not expanded in Emacs. The `dap--put-if-absent` in that case results in a wrong config, since it does not override it. For consistency, I used `plist-put` also for `projectDir`.

yyoncho: FYI workspaceRoot is expanded, we support that. I thought that users might wanna change cwd, this is common for other adapters. I will merge it as it is, later we can iterate on this implementation.

yyoncho:
Thank you!

dap-erlang.el
robertoaloi: Interesting, that wasn't working in my installation. I may be doing something wrong, I will double check.

PR #496, opened 2021-08-11
 dap-cpptools-setup set file mode on mac mono

darin-costello:
It currently checks if mono-mac exists, then attempts to set file modes on the linux mono


yyoncho (approved):

yyoncho:
Thank you

Issue #490, opened 2021-07-13
"Python :: Run pytest (at point)" does not find test method nested in test class


TauPan:
The message is always `‘dap-python’: no test at point`.

The problem seems to be in this code https://github.com/emacs-lsp/dap-mode/blob/2354d559a503fa992fe1d83566102dec62bedd0f/dap-python.el#L143 using mapcar which only finds the test classes.

pytest even allows arbitrarily nested test classes.

I don't know dash.el or lsp so I'm afraid I have no idea how to fix this.

Consider this example `test.py`:
~~~python
class TestClass:

    def test_method(self):
        assert True


def test_function():
    assert True
~~~

Running the template in `test_method` will find no test, while it will work inside `test_function`.

martin-liu:
I found the same problem when trying dap python, after some check, looks like the issue is `dap-python--parse-lsp-symbol` didn't process `children` which returned from `lsp--get-document-symbols` .

```elisp
(defun dap-python--test-at-point ()
  (->> (lsp--get-document-symbols)
       (mapcar #'dap-python--parse-lsp-symbol)  ; <-- here
       (dap-python--symbols-before-point (dap-python--cursor-position))
       dap-python--nearest-test))
```

`(lsp--get-document-symbols)` will return something like below, you can see there's `children` which have info for `test_method`:
```elisp
(#s(hash-table size 6 test equal rehash-size 1.5 rehash-threshold 0.8125 data
               ("name" "TestClass"
                "kind" 5
                "deprecated" nil
                "range" #s(hash-table size 2 test equal rehash-size 1.5 rehash-threshold 0.8125 data ("start" #s(hash-table size 2 test equal rehash-size 1.5 rehash-threshold 0.8125 data ("line" 0 "character" 0))
                                                                                                      "end" #s(hash-table size 2 test equal rehash-size 1.5 rehash-threshold 0.8125 data ("line" 3 "character" 19))))
                "selectionRange" #s(hash-table size 2 test equal rehash-size 1.5 rehash-threshold 0.8125 data
                                               ("start" #s(hash-table size 2 test equal rehash-size 1.5 rehash-threshold 0.8125 data ("line" 0 "character" 6))
                                                "end" #s(hash-table size 2 test equal rehash-size 1.5 rehash-threshold 0.8125 data ("line" 0 "character" 15))))
                "children" [#s(hash-table size 6 test equal rehash-size 1.5 rehash-threshold 0.8125 data
[...]

PR #503, opened 2021-09-07
ensure pytest at point works with class methods

martin-liu:
This will fix https://github.com/emacs-lsp/dap-mode/issues/490

`Python :: Run pytest (at point)` will work for class methods, e.g.
```python
class TestClass:

    def test_method(self):
        assert True
```

### DIFF
--- a/dap-python.el
+++ b/dap-python.el
@@ -88,7 +88,7 @@ https://github.com/pyenv/pyenv-which-ext."
 (lsp-defun dap-python--parse-lsp-symbol
   ((&DocumentSymbol
     :name :kind
-    :selection-range (&Range :start (&Position :line start-line
+    :range (&Range :start (&Position :line start-line
                                                :character start-character)
                              :end (&Position :line end-line
                                              :character end-character))))
@@ -113,7 +113,8 @@ https://github.com/pyenv/pyenv-which-ext."
 
 (defun dap-python--test-p (lsp-symbol)
   (let ((name (dap-python--symbol-name lsp-symbol)))
-    (and (string= (dap-python--symbol-type lsp-symbol) "Function")
+    (and (or (string= (dap-python--symbol-type lsp-symbol) "Function")
+             (string= (dap-python--symbol-type lsp-symbol) "Method"))
          (s-starts-with? "test_" name))))
 
 (defun dap-python--test-class-p (test-symbol lsp-symbol)
@@ -142,6 +143,7 @@ https://github.com/pyenv/pyenv-which-ext."
 
 (defun dap-python--test-at-point ()
   (->> (lsp--get-document-symbols)
+       (lsp--symbols->document-symbols-hierarchy)
        (mapcar #'dap-python--parse-lsp-symbol)
        (dap-python--symbols-before-point (dap-python--cursor-position))
        dap-python--nearest-test))


dap-python.el
```diff
@@ -88,7 +88,7 @@ https://github.com/pyenv/pyenv-which-ext."
 (lsp-defun dap-python--parse-lsp-symbol
   ((&DocumentSymbol
     :name :kind
-    :selection-range (&Range :start (&Position :line start-line
+    :range (&Range :start (&Position :line start-line
```
martin-liu, line 91: `selection-range` only have `start`-`end` range for `selection`, which cover the name of class/method.
`range` is what we need.

dap-python.el
```diff
@@ -113,7 +113,8 @@ https://github.com/pyenv/pyenv-which-ext."
 
 (defun dap-python--test-p (lsp-symbol)
   (let ((name (dap-python--symbol-name lsp-symbol)))
-    (and (string= (dap-python--symbol-type lsp-symbol) "Function")
+    (and (or (string= (dap-python--symbol-type lsp-symbol) "Function")
+             (string= (dap-python--symbol-type lsp-symbol) "Method"))
```
martin-liu, line 117: Class method will show `Method` in type. (kind 6)

dap-python.el
```diff
@@ -142,6 +143,7 @@ https://github.com/pyenv/pyenv-which-ext."
 
 (defun dap-python--test-at-point ()
   (->> (lsp--get-document-symbols)
+       (lsp--symbols->document-symbols-hierarchy)
```
martin-liu, line 146: This method will recursively add children to a flatten list, so that we'll have `(list parent child grandchild ..)`, i.e. the class methods will be in the list of simbols. 

```elisp
(defun lsp--document-symbols->document-symbols-hierarchy (document-symbols current-position)
  "Convert DOCUMENT-SYMBOLS to symbols hierarchy on CURRENT-POSITION."
  (-let (((symbol &as &DocumentSymbol? :children?)
          (seq-find (-lambda ((&DocumentSymbol :range))
                      (lsp-point-in-range? current-position range))
                    document-symbols)))
    (if children?
        (cons symbol (lsp--document-symbols->document-symbols-hierarchy children? current-position))
      (when symbol
        (list symbol)))))
  ```

martin-liu:
@yyoncho @nbfalcon Can you take a look?

yyoncho (approved):

yyoncho:
Looks good to me. Can you tell me which servers you have tested with? (It will be nice to cover pyright and pylsp)

martin-liu:
@yyoncho I have tested with pylsp & mspyls, not able to test pyright though. (I'm using Doom Emacs, but switch to pyright didn't work for me)

martin-liu:
@yyoncho Do you think this is good to merge? Let me know if there's any concern.

yyoncho:
Sorry, I forgot about this one. 

Thank you for contributing to dap-mode!

TauPan:
One minor quirk while testing this (with 5d1aa24a38bfc5dadedd6bd399791a550749e627 which arrived on melpa yesterday).

The test method must be named `test_<something>` for this to work... I usually name the first test method or function in a module/class simply `test` for lack of a better idea what I want to test (while exploring things), which is ok with pytest, and this promptly gave me the same error as before :-)

~~~elisp
(defun dap-python--test-p (lsp-symbol)
  (let ((name (dap-python--symbol-name lsp-symbol)))
    (and (or (string= (dap-python--symbol-type lsp-symbol) "Function")
             (string= (dap-python--symbol-type lsp-symbol) "Method"))
         (s-starts-with? "test_" name))))
~~~

see https://docs.pytest.org/en/6.2.x/goodpractices.html#conventions-for-python-test-discovery which is not really consistent with the example in https://docs.pytest.org/en/6.2.x/getting-started.html#group-multiple-tests-in-a-class which speaks of `test_` prefixed methods.

But this is not really important as my test methods or functions (and everybody elses, I guess) usually match the `test_*` glob.

PR #504, opened 2021-09-09
make dap-python--debug-test-at-point work

TauPan:
After #503 was merged I could run dap-debug and select the template, but binding `dap-python-debug-test-at-point` directly did not.

At first I got the error that the template is not a string and then that the templates doesn't have `:type` (which it did, it was just wrong and the error was misleading).

Fortunately I was able to fix those myself.

### DIFF
--- a/dap-python.el
+++ b/dap-python.el
@@ -152,7 +152,7 @@ https://github.com/pyenv/pyenv-which-ext."
   "Return the debug template whose name is TEMPLATE-NAME.
 For the name, only the template's `car' is checked, not its
 `:name' property."
-  (--first (string= template-name it) dap-debug-template-configurations))
+  (cdr (--first (string= template-name (car it)) dap-debug-template-configurations)))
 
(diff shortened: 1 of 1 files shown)


yyoncho:
@martin-liu willing to take a look?

martin-liu (approved):
LGTM.

Ideal case would be have unit test to cover this, but looks like we didn't have good tests.  

@TauPan Regarding [the `test_` issue that you mentioned](https://github.com/emacs-lsp/dap-mode/pull/503#issuecomment-915901833), the fix is just change prefix check from `test_` to `test`,  maybe we could update that as well?

TauPan:
> @TauPan Regarding [the `test_` issue that you mentioned](https://github.com/emacs-lsp/dap-mode/pull/503#issuecomment-915901833), the fix is just change prefix check from `test_` to `test`, maybe we could update that as well?


I've created #505 to remind myself.

PR #507, opened 2021-09-13
drop '_' from test function/method prefix

### LINKED ISSUE #505: Use prefix "test" instead of "test_" for matching test methods.
One minor quirk while testing #503 (with 5d1aa24a38bfc5dadedd6bd399791a550749e627 which arrived on melpa yesterday).

The test method must be named `test_<something>` for this to work... I usually name the first test method or function in a module/class simply `test` for lack of a better idea what I want to test (while exploring things), which is ok with pytest, and this promptly gave me the same error as before :-)

~~~elisp
(defun dap-python--test-p (lsp-symbol)
  (let ((name (dap-python--symbol-name lsp-symbol)))
    (and (or (string= (dap-python--symbol-type lsp-symbol) "Function")
             (string= (dap-python--symbol-type lsp-symbol) "Method"))
         (s-starts-with? "test_" name))))
~~~

see https://docs.pytest.org/en/6.2.x/goodpractices.html#conventions-for-python-test-discovery which is not really consistent with the example in https://docs.pytest.org/en/6.2.x/getting-started.html#group-multiple-tests-in-a-class which speaks of `test_` prefixed methods.

But this is not really important as my test methods or functions (and everybody elses, I guess) usually match the `test_*` glob.

_Originally posted by @TauPan in https://github.com/emacs-lsp/dap-mode/issues/503#issuecomment-915901833_

TauPan:
fixes #505

### DIFF
--- a/dap-python.el
+++ b/dap-python.el
@@ -115,7 +115,7 @@ https://github.com/pyenv/pyenv-which-ext."
   (let ((name (dap-python--symbol-name lsp-symbol)))
     (and (or (string= (dap-python--symbol-type lsp-symbol) "Function")
              (string= (dap-python--symbol-type lsp-symbol) "Method"))
-         (s-starts-with? "test_" name))))
+         (s-starts-with? "test" name))))
 
 (defun dap-python--test-class-p (test-symbol lsp-symbol)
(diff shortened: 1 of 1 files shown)


yyoncho:
Thank you!

PR #508, opened 2021-09-20
improve dap-netcore out of the box experience

s-kostyaev:
Hi. Thanks for this cool project!

I have spent some time to figure out how to use dap-netcore. After initial setup it works flawlessly. Maybe this PR can help someone to spare some time. 
P.S. I also think it's worth it to add `dap-netcore` into documentation. But I think it's the task for another PR.

### DIFF
--- a/dap-netcore.el
+++ b/dap-netcore.el
@@ -57,7 +57,8 @@ Will be set automatically in Emacs 27.1 or newer with libxml2 support."
     (if (and (not dap-netcore-download-url)
 	     (fboundp 'libxml-available-p)
 	     (libxml-available-p)
-	     (fboundp 'dom-search))
+	     (fboundp 'dom-search)
+	     (fboundp 'dom-attr))
 	(url-retrieve "https://github.com/Samsung/netcoredbg/releases"
 		      (lambda (_)
 			(setq dap-netcore-download-url
--- a/dap-netcore.el
+++ b/dap-netcore.el
@@ -29,9 +29,8 @@
 (require 'dap-mode)
 (require 'f)
 
-(defcustom dap-netcore-install-dir (f-slash (f-join user-emacs-directory ".cache" "lsp" "netcoredbg"))
-  "Install directory for netcoredbg.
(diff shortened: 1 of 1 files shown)


dap-netcore.el
```diff
@@ -85,6 +85,24 @@ The slash is expected at the end."
     ("attach"
      (dap--put-if-absent conf :processId (string-to-number (read-string "Enter PID: " "2345"))))))
 
+(if (and (fboundp 'libxml-available-p)
```
kiennq, line 88: Why do you need this? Is the original `dap-netcore-download-url` link not working?
Having this code here will make a very perceivable delay when the user is loading `dap-netcore`, I don't really recommend it.

s-kostyaev: Yes, original link is broken

s-kostyaev: Also in modern emacs this code run asynchronously. I can't see any noticeable delay. But I use emacs 28 with native-compilation. Could someone test it on old emacs?

kiennq, line 88: `when`  instead of `if` for if clause without else part

s-kostyaev: fixed, thanks

kiennq: In that case I think you should change the `dap-netcore-download-url` (both description and default value) to include `nil` as let `dap-netcore` to decide where to download. Then move these codes inside `dap-netcore--debugger-install` so that it will not affect loading time.

kiennq: Also, you should probably consider to use `lsp-download-install` in `dap-netcore--debugger-install` as well to prevent Emacs hangs while waiting for download

kiennq: @s-kostyaev

s-kostyaev: Will see, thank you.

s-kostyaev: done. @kiennq

yyoncho (approved):

jcs090218 (approved):

PR #513, opened 2021-09-26
dap-ui: Fixed issue where control icon background was unspecified

### LINKED ISSUE #462: dap-ui-controls icons do not show when :background attribute of fringe face is not set.
At the moment the icons of the dap-ui-controls posframe popup do not show up if the
`:background` property of the `fringe` face is unset.

`dap-ui--create-command` constructs the image property for the icons like so
```lisp
 (image :type png
                               :file ,(f-join dap-ui--control-images-root-dir image)
                               :ascent center
                               :background ,(face-attribute 'fringe :backgound nil t))
```
Ideally there would be another fallback or documentation telling the user to set the face attribute.

adammohammed:
This resolves #462.

If a user starts `dap-debug`, dap-mode creates a set of icons used to do common debug actions (step, continue, etc). These icons are created using `propertize` to make the text that displays the images on a background.

The existing setting uses the background color from the `fringe` face to set the background of the icon. The return value for the `face-attribute` function is not guaranteed to be specified. When the `:background` is unspecified, the space is made for the icons but nothing is displayed.

This PR defines a face to control the background color used for the debug icons. It defaults to using the background color defined in the `fringe` face, but will inherit from the `default` face if the `:background` attribute is unspecified in `fringe`.

### DIFF
--- a/dap-ui.el
+++ b/dap-ui.el
@@ -151,6 +151,11 @@ number - expand N levels."
   "Face for enabled breakpoint icon in fringe."
   :group 'dap-ui)
 
+(defface dap-ui-controls-fringe
+  '((t :inherit fringe))
+  "Face used for the background of debugger icons in fringe."
+  :group 'dap-ui)
+
 (defcustom  dap-ui-default-fetch-count 30
   "Default number of variables to load in inspect variables view for
 array variables."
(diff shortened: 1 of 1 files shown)


yyoncho (approved):

yyoncho:
Thank you!

PR #516, opened 2021-10-03
dap-ui: Display `instructionPointerReference` when available

eshelyaron:
When provided by the debug adapter server, add the instruction pointer of each frame alongside its name in the `*dap-ui-sessions*` buffer.

For instance, the [SWI-Prolog debug adapter](https://github.com/eshelyaron/debug_adapter) reports for each frame exactly in which instruction pointer or stage the frame is stopped.

Here's an example of how the sessions tree looks with this addition:
Screen Shot 2021-10-03 at 11 44 05

### DIFF
--- a/dap-ui.el
+++ b/dap-ui.el
@@ -698,7 +698,7 @@ DEBUG-SESSION is the debug session triggering the event."
                             (funcall
                              callback
                              (-map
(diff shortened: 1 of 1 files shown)


yyoncho (approved):

yyoncho:
Thank you!

eshelyaron:
With pleasure, thanks for `dap-mode`!

PR #529, opened 2021-10-20
Fix dap--put-if-absent to allow unsetting keys

davidarnold:
Currently `dap--put-if-absent` uses the truthiness of the value returned by `plist-get` to determine whether a key is present, which seems incorrect.  This PR changes it to use `plist-member` to test for key presence.

**Example:**

The `dap-elixir` provider sets `taskArgs` and `requireFiles` in the default debug template, which causes failures when trying to run a `task` of `phx.server` instead of `test`.  The current logic of `dap--put-if-absent` makes it impossible to override those keys to `nil` in your local template.  So instead, you have to concoct a truthy Emacs Lisp value that is a no-op at the `mix` level.

Expected working template:
```elisp
(dap-register-debug-template
 "Phoenix Server"
 (list :type "Elixir"
       :name "mix phx.server"
       :task "phx.server"
       :taskArgs nil
       :requireFiles nil))
```

Actual working template:
```elisp
(dap-register-debug-template
 "Phoenix Server"
 (list :type "Elixir"
       :name "mix phx.server"
       :task "phx.server"
       :taskArgs '(nil)
       :requireFiles '("fake")))
```

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -377,7 +377,7 @@ has succeeded."
 
 (defmacro dap--put-if-absent (config key form)
   "Update KEY to FORM if KEY does not exist in plist CONFIG."
-  `(if (plist-member ,config ,key) ,config (plist-put ,config ,key ,form)))
+  `(let ((c ,config)) (if (plist-member c ,key) c (plist-put c ,key ,form))))
 
 (defun dap--completing-read (prompt collection transform-fn &optional predicate
                                     require-match initial-input


davidarnold:
@yyoncho not sure if you're already working on it, but the bytecode overflow seems to have the following characteristics:

- It's reproducible on master (so, unrelated to this PR)
- It's fixable by removing any of the `dap--put-if-absent` lines in `dap-elixir.el`
- It's reproducible in `dap-gdb--lldb.el` by adding an extra `dap--put-if-absent` line (any key and value) to the `dap-gdb-lldb--populate-gdbserver` function

It seems strictly related to the number of `dap--put-if-absent` calls inside of the `->` block.

- 6 calls compiles immediately
- 7 calls is noticeably slower
- 8 calls is extremely slow
- 9 calls eventually fails with an overflow

It's fixable by changing `dap--put-if-absent` to a function, so I'm guessing it's caused by the `->` macro operating recursively on the inner macro.  I almost changed `dap--put-if-absent` to a function in the PR since I didn't understand why it was a macro to begin with, but I wasn't sure if there was a good reason that I was missing.  I can add that change to this branch if you would like.

schellj:
@davidarnold @yyoncho note that this change has broken some of the default templates, such as the `Go Launch File Configuration` from [dap-go.el](https://github.com/emacs-lsp/dap-mode/dap-go.el). I.e., that template sets `:program nil` and there is logic (such as at https://github.com/emacs-lsp/dap-mode/blob/master/dap-go.el#L87) that uses `dap--put-if-absent` to set some of the config values (such as `:program`). The default `Go Launch File Configuration` template currently doesn't work as-is because the vscode extension is expecting `program` to be set, and with this change it does not get set.

davidarnold:
@schellj Seems easy enough to fix by removing all the unnecessary `nil` lines (not just `:program`) in the specific templates.  I don't use Go, so I can't test.  Can you give it a shot?

schellj:
@davidarnold that does work. My point was more that there are 15 or 20 `dap-*.el` packages that all may need to be updated.

davidarnold:
That would ultimately be a decision for @yyoncho but given the README,

> The API considered unstable until 1.0 release is out.

I would still advocate fixing this underlying functionality, and also follow up with fixing the out of the box templates.  If we revert this change, then there is no way to remove a template key that is specified by the language's `*--populate-default-args`.  That would necessitate making the default args much more conservative (i.e. removing many keys) which would also result in breakages for custom configs.

yyoncho:
I think that at this point the easiest solution which will possibly make everyone happy is to alter @davidarnold to use artificial value (e. g. `:dap-nil`) which will stand for `nil` that you don't want to replace.

davidarnold:
@yyoncho Do you mean reverting this change and implementing some other logic in `dap--put-if-absent` that specifically looks for a magic symbol?

I'm predicting that it will be difficult for users to find and use that magic symbol instead of doing the more obvious thing of setting `nil` in their templates.  Personally, I spent a few hours in frustration debugging this behavior only to find out that setting `nil` in my templates gets ignored due to a specious use of the `plist` library.

Certainly the pattern has already been established, but I can't understand the widespread usage in the templates of `nil` as a no-op.  Phrasing it back, it would be like saying "I absolutely want this line to be present in my code and I absolutely want it to do nothing."  Seems odd, no?

I still believe that this change in this PR (coupled with additional work to fix the default templates) will be best for the long term usability of the project.  However, if you feel that version 0.7 is too mature to have a breaking change, I will revert my work.  I don't want to cause damage to the project or its users.

yyoncho:
@davidarnold I agree with your concerns they make sense. Being backward compatible is not my main concern here but it is how "coupled with additional work to fix the default templates" goes. 

> @yyoncho Do you mean reverting this change and implementing some other logic in dap--put-if-absent that specifically looks for a magic symbol?

Yes and no. We can have the special value to mean `:replace-me` instead of `:skip-me`. Note that we should keep `dap-debug-edit-template` functional.

marun:
I attempted to configure dap-go for the first time and ran into this issue of `:program nil` no longer being set on invocation. Pardon my naivety, but might this be an opportunity to add unit testing to avoid future regression? If so, I'd be happy to contribute (with appropriate guidance of course).

yyoncho:
I reverted the PR until we find a solution.

Issue #543, opened 2021-12-14
dap-debug fails immediately with (wrong-type-argument stringp nil) 


dbactual:
I am also having the same problem as bugs #7, #31 and #479. This is the backtrace:

```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  make-network-process(:name "Run Configuration" :buffer nil :host "127.0.0.1" :service nil :nowait nil :tls-parameters nil)
  open-network-stream("Run Configuration" nil "127.0.0.1" nil :type plain :coding no-conversion)
  dap--open-network-stream("Run Configuration" "127.0.0.1" nil)
  dap--create-session((:type "remote" :request "launch" :args "" :name "Run Configuration" :debugPort 8888 :host "127.0.0.1"))
  dap-start-debugging-noexpand((:type "remote" :request "launch" :args "" :name "Run Configuration" :debugPort 8888 :host "127.0.0.1"))
  #f(compiled-function () #<bytecode 0x1fe95469cb09>)()
  dap-debug((:type "remote" :request "launch" :args "" :name "Run Configuration"))
  funcall-interactively(dap-debug (:type "remote" :request "launch" :args "" :name "Run Configuration"))
  call-interactively(dap-debug record nil)
  command-execute(dap-debug record)
  execute-extended-command(nil "dap-debug" "dap-deb")
  funcall-interactively(execute-extended-command nil "dap-debug" "dap-deb")
  call-interactively(execute-extended-command nil nil)
  command-execute(execute-extended-command)
```

Here is my .emacs setup:

```
(dap-register-debug-provider
 "remote"
 (lambda (conf)
   (plist-put conf :debugPort 8888)
   (plist-put conf :host "127.0.0.1")
   conf))

(dap-register-debug-template "remote"
                             (list :type "remote"
                                   :request "launch"
                                   :args ""
                                   :name "Run Configuration"))
```

Versions:

```
dap-mode           20211117.1555
lsp-mode           20211214.1007
GNU Emacs 27.2 (build 1, x86_64-apple-darwin20.3.0) of 2021-03-25
```

Is it suspicious that `open-network-stream` makes no mention of the port?

Many thanks for any hints or tips here!

dbactual:
An update: looks like an argument name was a problem:

```
diff --git a/dap-mode.el b/dap-mode.el
index 05efbfd..ad09c86 100644
--- a/dap-mode.el
+++ b/dap-mode.el
[...]

PR #552, opened 2022-01-26
fix argument ref in (dap--create-session)

parsoj:
I'm seeing an issue while trying to set up a new/custom dap provider, where the `debugPort` argument referenced in the docs:
https://github.com/emacs-lsp/dap-mode/blob/76cad34de8984f57c2b1e374e9c985cc7ec8dad0/docs/page/adding-debug-server.md?plain=1#L23

 isn't being read in on the call to `(dap--open-network-stream)`
 
 It appears that the argument that sets the port is actually called `debugServer`.  I think this conflicts with the docs, and is maybe a bit of a confusing naming. 
 
 Digging through the issue history it looks like some others have run into the same thing and a similar fix worked for them:
 https://github.com/emacs-lsp/dap-mode/issues/543
 
 But I haven't seen a PR up for this, yet - so I figured I'd get the ball rolling!


eshelyaron:
I've just hit this issue as well, I thought of putting together a PR as this is quite a "gotcha", and was glad to see someone already did :)

One thing to note though is that a backward-compatible solution may be preferable. A quick GitHub search shows at least [one more project](https://github.com/fracek/cairo-dap/blob/6df0c4e56ca613b24b33ad8851ee7b27f650a4d8/editors/emacs/dap-cairo.el#L14) that is already using `debugServer`.
Maybe accept both options and prefer `debugPort`?

PR #561, opened 2022-02-16
Added configuration for Unity debug adapter

23Suvu:
Adding support for Unity (game engine) debug adapter.

Known Issues :
* Adapter will often fail to download from the vs marketplace requiring manual intervention.
* The adapter attempts to use DAP events that are not yet configured in dap-mode. This results in some "missing event" spam in the Messages buffer. The adapter is still very much usable, and the missing events can still be addressed in the future.


yyoncho:
Looks good. Can you drop a line in the readme that we support it?

yyoncho:
Thank you!

PR #562, opened 2022-02-17
Add dap-debug-restart to dap-hydra

samhedin:
This was discussed on Discord. The existing debug commands in `dap-hydra` do not kill the current debugging sessions, which leads to them piling up if the user isn't aware. In my case, it made emacs fill up my memory and crash.
The command that should be used if the user isn't interested in the old sessions is `dap-debug-restart`. This PR adds it to the hydra. `dr` was already taken, so I arbitrarily set it to `ds`. This can, of course, be changed. The same goes for the description. 
Maybe the description should contain some information that this is usually what you want if you're inside a debug session? I don't know how to express it succinctly though.


yyoncho (approved):

yyoncho:
Thank you!

Issue #440, opened 2021-02-19
Download links for dap-netcore not correct


samwdp:
The dowload links for all versions of the netcoredbg package on github are incorrect. The latest versions are:
https://github.com/Samsung/netcoredbg/releases/download/1.2.0-672/netcoredbg-linux-bionic-amd64.tar.gz
https://github.com/Samsung/netcoredbg/releases/download/1.2.0-672/netcoredbg-linux-focal-amd64.tar.gz
https://github.com/Samsung/netcoredbg/releases/download/1.2.0-672/netcoredbg-osx.tar.gz
https://github.com/Samsung/netcoredbg/releases/download/1.2.0-672/netcoredbg-win64.zip

josteink:
I'm able to install the netcoredbg DAP server manually, but after doing that I'm getting the same error when trying to debug using the "launch" option.

Specifically I'm noticing the error says "You intended to execute a .NET program, but **dotnet-**/the/actual/path/to/the/binary does not exist."

Notice the `dotnet-` prefix. If that is what someone (the debugger?) is looking for, it's obviously not going to work. Where does that prefix come from? Anyway to get it removed?

I managed to get debugging working by manually launching the program and attaching to a PID, but that's quite cumbersome.

samwdp:
@josteink I have also now managed to set up debugging when launcing the program and attaching it to a PID. I cannot for the life of me get a launch option working

josteink:
I have a feeling there is something super-simple that I’m missing (invoke through main dotnet CLI?), but I have no idea what that would be. 

Can someone who has made this work please help us with a 100% step-by-step, idiot proof way of launching into debug-mode?

samwdp:
I have the "attach" working. I dont know if you are using a distribution or have your own config but I have this setup in mine
```elisp
 (use-package dap-mode
    :commands dap-debug
    :hook (dap-mode . dap-tooltip-mode)
    :config
    (require 'dap-node)
    (require 'dap-netcore)
    (require 'dap-lldb)
    (require 'dap-cpptools)
    (setq dap-netcore-install-dir "c:/tools/"))
```
I have hade to change the install dir as every time I open the debugger, it cannot find the netcoredbg.exe in the internal location ~/.emacs.d/.cache/lsp/netcoredbg

yyoncho:
https://github.com/emacs-lsp/dap-mode/pull/564

PR #566, opened 2022-02-19
Fix error intstalling netcoredb on GNU/Linux

josteink:
Fix runtime error. Add missing parameter to `string-match-p` clause.

This PR fixes an error/regression introduced in https://github.com/emacs-lsp/dap-mode/issues/440 for GNU/Linux users.

### DIFF
--- a/dap-netcore.el
+++ b/dap-netcore.el
@@ -80,7 +80,7 @@ Will be set automatically in Emacs 27.1 or newer with libxml2 support."
 				 (libxml-parse-html-region (point-min) (point-max))
 				 (lambda (node)
 				   (string-match-p (pcase system-type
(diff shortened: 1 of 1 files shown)


yyoncho:
cc @s-kostyaev

s-kostyaev (changes requested):
My bad. But it better to use here `system-configuration` variable

josteink:
Fixed

s-kostyaev:
LGTM, thank you

s-kostyaev:
@yyoncho we can merge it

yyoncho:
Thank you

josteink:
Thanks for the quick merge 🙂

PR #576, opened 2022-03-06
Repaired debugpy to work with flask

DJManas:
Hello,

I had found the problem and repaired it for debugpy and flask. Maybee not the cleanest, but I am new to LISP and I can finally debug my app to move with my project.

Thanks,
Regards,
Petr Sourek


DJManas:
Please do not approve, it has a bug in single debug, which I will repair with next commit. Thanks

DJManas:
Now it should work.

DJManas:
But I had found another issue. When I use PTVSD it doesn't take into account :env variables for first time. It ends with error, then I use dap-debug-last and its working flawlesly, but my LISP Fu is not strong enough to pinpoint the problem. Should I create new bug for someone to take a look at?

PR #578, opened 2022-03-08
Add support for VSCode Tasks

elken:
# Purpose

The idea of this PR was initially to just handle [`preLaunchTask`](https://code.visualstudio.com/docs/editor/debugging#_launchjson-attributes), which is a more robust, vscode-native version of `dap-compilation` for running tasks before a debug configuration. It shortly spiraled into the current PR, which I don't think goes _too_ far in terms of support.

It aims to be much more in line with vscode, and should be able to handle many more setups ootb.

# Changes

## Move `sanitize-json` to a utils function

Straightforward, it's used in `dap-tasks` and `dap-launch`.

## Add support for handling `preLaunchTask`

Correctly handles tasks specified in a debug config's `preLaunchTask`. If no task is found matching the name, the config errors (much as it would in vscode). Down the line, `dap-compilation` could potentially be removed; but as it stands currently it actually takes precedence (so as to not break current setups). This behaviour could be gated behind a setting, if required.

## Add support for `dependsOn` 

Allows a hierarchy of tasks to be created (to say build several dependent projects first, then debugging this one).

# Summary

Open to any and all critique, hopefully we can get this merged in a good state :D

I did try and also respect [`presentation`](https://code.visualstudio.com/Docs/editor/tasks#_output-behavior) options, but I couldn't quite work it. `save-window-excursion` solves the problem, but I couldn't make it work in a toggleable way.

### DIFF
--- a/dap-launch.el
+++ b/dap-launch.el
@@ -26,27 +26,6 @@
 (require 'lsp-mode)
 (require 'json)
 
-(defun dap-launch-sanitize-json ()
-  "Remove all C-style comments and trailing commas in the current buffer.
-Comments in strings are ignored. The buffer is modified in place.
-Replacement starts at point, and strings before it are ignored,
-so you may want to move point to `point-min' with `goto-char'
-first. This function moves `point'. Both // and /**/ comments are
-supported."
-  (while (re-search-forward
-          (rx
-           (or (group
-                (or (: "//" (* nonl) eol)
-                    (: "/*" (* (or (not (any ?*))
-                                   (: (+ ?*) (not (any ?/))))) (+ ?*) ?/)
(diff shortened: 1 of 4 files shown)


yyoncho:
can you resolve conflicts?

yyoncho (approved):

yyoncho:
I approved it, lets's iterate based on the feedback.

elken:
Oh conflicts, good catch! I'll look at these now :D

elken:
@yyoncho can you review again please? I'm not sure about my merge conflict resolution

elken:
Paging Dr. @yyoncho :)

yyoncho (approved):

PR #590, opened 2022-04-04
python: make test-at-point work

stfl:
The debug-template for pytest (at point) now has the correct :type "python-test-at-point"
The corresponding debug-provider dap-python--populate-test-at-point is loaded


stfl:
I just came across the problem again and this PR still fixes it.

stfl:
my workaround within doom-emacs

```lisp
(after! (dap-python)
  (dap-register-debug-template "Python :: Run pytest (at point)"
                               (list :type "python-test-at-point"
                                     :args ""
                                     :program nil
                                     :module "pytest"
                                     :request "launch"
                                     :name "Python :: Run pytest (at point)")))
```

PR #593, opened 2022-04-10
Bare minimum Kotlin Debug Adapter setup

themkat:
The absolute minimum configuration for using the [Kotlin Debug Adapter](https://github.com/fwcd/kotlin-debug-adapter) to debug Kotlin code. For some reason, [lsp-kotlin already has a variable for debug adapter path](https://github.com/emacs-lsp/lsp-mode/blob/master/clients/lsp-kotlin.el#L80), so reusing that one.

Currently only launch-requests are supported, which is a huge step forward from nothing. Future work may target attach requests as well, but I never use that, so would require some additional research from me to implement.

### DIFF
--- a/dap-kotlin.el
+++ b/dap-kotlin.el
@@ -1,3 +1,6 @@
+;; Commentary:
+;; Adapter for https://github.com/fwcd/kotlin-debug-adapter
+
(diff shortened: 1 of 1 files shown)


yyoncho:
Thank you!

PR #599, opened 2022-04-20
Add dap-ui-breakpoints-browse

samhedin:
Allows the user to view and jump to breakpoints using completing-read

This was discussed on Discord. @yyoncho also suggested encoding the state of breakpoints with colors, which I'm all for. But I'm not sure how it should look, so taking suggestions there. Just a colored dot at the end of the string, maybe? Not familiar with how this would be done, but I guess that `dap-ui-breakpoints` is a good starting point.
Screenshot from 2022-04-20 21-55-51

There are some additions I want personally, such as using consult-preview to view the entire context of the breakpoint, as well as the ability to delete them with embark. However, that might be better suited for my personal config.

### DIFF
--- a/dap-ui.el
+++ b/dap-ui.el
@@ -447,6 +447,31 @@ DEBUG-SESSION is the debug session triggering the event."
     (-some-> existing-breakpoint (plist-get :marker) (set-marker nil))
     (dap--breakpoints-changed (cl-remove existing-breakpoint file-breakpoints) file-name)))
 
+(defun dap-ui-breakpoints-browse ()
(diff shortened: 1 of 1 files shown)


samhedin:
To add the icon, I switched to using `dap-ui--breakpoints-data` rather than `dap--get-breakpoints`, as it included all the icon information. This is against Discord recommendation so it can be switched back if needed, it just seemed like the natural thing to do..

yyoncho (approved):

yyoncho:
Thank you!

Issue #604, opened 2022-04-23
[dap-go] how to debug specify test function


edenzhong7:
It seems that debug go test function will run all tests in current package, could I only debug single one test function like vscode? I saw that vscode launchs test bin with extra parameters ` -test.run ^TestLogger$ `.

edenzhong7:
another question is how to show the doc of var or function in posframe instead of minibuffer, currently only can see the function signature without doc

yyoncho:
@s-kostyaev is using go, so he might be interested in implementing this feature. As for the second question, you may use lsp-ui mouse hover or set lsp-ui-doc-show-with-cursor to t. Refer to https://emacs-lsp.github.io/lsp-mode/tutorials/how-to-turn-off/

edenzhong7:
> @s-kostyaev is using go, so he might be interested in implementing this feature. As for the second question, you may use lsp-ui mouse hover or set lsp-ui-doc-show-with-cursor to t. Refer to https://emacs-lsp.github.io/lsp-mode/tutorials/how-to-turn-off/

thanks a lot, let me take a try

s-kostyaev:
Implemented. See docs - https://emacs-lsp.github.io/dap-mode/page/configuration/#go

kprav33n:
I see that this has been implemented only in `dap-dlv-go`. Is it possible to get the same behavior in `dap-go`?

s-kostyaev:
> I see that this has been implemented only in `dap-dlv-go`. Is it possible to get the same behavior in `dap-go`? The original issue was for `dap-go`.

I think it is possible. But `dap-go` doesn't work for me. So I can't check if new code will work.

yyoncho:
@kprav33n consider migrating to `dap-dlv-go`. dap-go will be deprecated at some point.

PR #611, opened 2022-05-02
Add go test debug current function

s-kostyaev:
Add ability to debug current test function.

### DIFF
--- a/dap-dlv-go.el
+++ b/dap-dlv-go.el
@@ -30,6 +30,8 @@
 (require 'dap-mode)
 (require 'dap-utils)
 (require 'f)
+(require 'lsp-mode)
+(require 'dash)
 
 (defcustom dap-dlv-go-delve-path
   (or (executable-find "dlv")
@@ -50,6 +52,8 @@
   (setq conf (pcase (plist-get conf :mode)
                ("auto"
 		(dap-dlv-go--populate-auto-args conf))
+	       ("test"
+		(dap-dlv-go--populate-test-args conf))
                ("debug"
 		(dap--put-if-absent
 		 conf :program (f-dirname (buffer-file-name))))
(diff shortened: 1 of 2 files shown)


s-kostyaev:
fixes #604

s-kostyaev:
@edenzhong7 try if it works for you. You should add breakpoint, then `M-x` `dap-debug` `<RET>` `Go Dlv Test Current Function Configuration` `<RET>`

dap-dlv-go.el
```diff
@@ -93,6 +99,18 @@
       (plist-put conf :mode "test")
     (plist-put conf :mode "debug")))
 
+(defun dap-dlv-go--populate-test-args (conf)
+  "Populate auto arguments according to CONF."
+  (dap--put-if-absent conf :program "."))
+
+(defun dap-dlv-go--extract-current-function-name ()
```
yyoncho, line 106: The alternative is to use `textDocument/symbols` similar to what dap-java.el do. I would say that it will return better results.

s-kostyaev: Updated

yyoncho (approved):

edenzhong7:
> 
Unfortunately, I just tried it but it doesn't work, error message:
`Warning (emacs): Initialize request failed: unrecognized request Disable showing Disable logging`

`(node:97628) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
`

s-kostyaev:
@edenzhong7 let's try from the beginning. In your error message there is `node`,  but for new debug adapter node not needed. Change your configuration according to https://emacs-lsp.github.io/dap-mode/page/configuration/#go

PR #616, opened 2022-05-06
dap-cpptools: Update to v1.9.8

abougouffa:
- Update the default version for `cpptools` to the latest
- Fix the path of the `OpenDebugAD7` binary in new version, it has changed from `extension/debugAdapters/OpenDebugAD7` to`extension/debugAdapters/bin/OpenDebugAD7` 
- Make new binaries executable

### DIFF
--- a/dap-cpptools.el
+++ b/dap-cpptools.el
@@ -32,13 +32,13 @@
   :group 'dap-cpptools
   :type 'string)
 
-(defcustom dap-cpptools-extension-version "0.29.0"
+(defcustom dap-cpptools-extension-version "1.9.8"
   "The version of the cpptools vscode extension."
   :group 'dap-cpptools
   :type 'string)
 
 (defcustom dap-cpptools-download-url
-  (format "https://github.com/microsoft/vscode-cpptools/releases/download/%s/cpptools-%s.vsix"
+  (format "https://github.com/microsoft/vscode-cpptools/releases/download/v%s/cpptools-%s.vsix"
           dap-cpptools-extension-version
           (alist-get system-type
                      '((windows-nt . "win32")
@@ -49,10 +49,8 @@
   :type 'string)
 
 (defcustom dap-cpptools-debug-program
-  `(,(concat dap-cpptools-debug-path
-             (if (eq system-type 'windows-nt)
-                 "/extension/debugAdapters/bin/OpenDebugAD7.exe"
-               "/extension/debugAdapters/OpenDebugAD7")))
+  `(,(concat dap-cpptools-debug-path "/extension/debugAdapters/bin/OpenDebugAD7"
+             (if (eq system-type 'windows-nt) ".exe" "")))
   "The path to the cpptools debug adapter."
   :group 'dap-cpptools
   :type '(repeat string))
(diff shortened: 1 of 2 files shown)


dap-cpptools.el
```diff
@@ -64,16 +62,23 @@ With prefix, FORCED to redownload the extension."
   (unless (and (not forced) (file-exists-p dap-cpptools-debug-path))
     (dap-utils--get-extension dap-cpptools-download-url dap-cpptools-debug-path)
     (let* ((adapter-binary (cl-first dap-cpptools-debug-program))
-           (mono (f-join (f-parent adapter-binary) "mono.linux-x86_64"))
-           (mono-mac (f-join (f-parent adapter-binary) "mono.osx"))
-           (lldb-mi (f-join (f-parent adapter-binary) "lldb-mi/bin/lldb-mi")))
+           (adapters-path (f-parent (f-parent adapter-binary)))
+           (extension-bins-path (f-join (f-parent adapters-path) "bin"))
+           (bins
+            (append (mapcar #'(lambda (path) (f-join extension-bins-path path))
```
yyoncho, line 68: nit: there is no longer need to `#'` before lambdas

yyoncho (commented):
Looks good. I wish we somehow warn the user if it has the older version and it does not have the new one. Otherwise, it will simply stop working for those using the current setup after upgrading it.

yyoncho:
> wish we somehow warn the user if it has the older version and it does not have the new one.

Actually, you have just refactored the code a little bit and the changes are only related to the download code. Am I reading it right?

abougouffa:
@yyoncho That's right! I will try to do something about it! Thank you for the review!

abougouffa:
> Actually, you have just refactored the code a little bit and the changes are only related to the download code. Am I reading it right?

No, it won't work after upgrading `dap-mode` and working with the old downloaded extension, as the path of `OpenDebugAD7` has changed.

> - Fix the path of the `OpenDebugAD7` binary in new version, it has changed from `extension/debugAdapters/OpenDebugAD7` to`extension/debugAdapters/bin/OpenDebugAD7` 

I'm working on an improvement, I added a function to parse the installed package's manifest and check the version. Next, I will check for it and propose to remove the installed extension and install the new one!

abougouffa:
@yyoncho I added a warning like you suggested. When a user with an older version of the extension loads the package, it prints a warning message saying:
```
Warning (emacs): You have an old cpptools v0.29.0. Please run a forced ‘dap-cpptools-setup’ to install the new v1.9.8. Run ‘M-x eval-expression’ then ‘(dap-cpptools-setup t)’
```

yyoncho:
> Run ‘M-x eval-expression’ then ‘(dap-cpptools-setup t)’

It should be `C-u M-x dap-cpptools-setup`

abougouffa:
> It should be `C-u M-x dap-cpptools-setup`

Yep, it will be better like this, however, I think you mean `C-u 1 M-x dap-cpptools-setup`

abougouffa:
@yyoncho I forced a push that fixes it.

Now the warning message will be:

```
Warning (emacs): You have an old cpptools v0.29.0. Please run ‘C-u 1 M-x dap-cpptools-setup’ to install the new v1.9.89.
```

abougouffa:
## Summery

- [x] Upgrade from cpptools 0.29.0 to 1.9.8, 
 - set file modes for new files in the extension, 
 - change the `OpenDebugAD7` directory,
 - refactor some related stuff.
- [x] Add a function `dap-utils-vscode-get-installed-extension-version` in `dap-utils.el` to check for a package installed in `path`, parse its manifest and extract the installed version.
- [x] Add a warning when loading `dap-cpptools` to warn users who have an older version installed, and to suggest forcing an upgrade by running `C-u 1 M-x dap-cpptools-setup`.

@yyoncho  Can you check that please?

dap-cpptools.el
yyoncho: can you apply this as well?

dap-utils.el
```diff
@@ -85,6 +85,18 @@ PATH is the download destination path."
                    (f-join dap-utils-extension-path "github" (concat owner "." repo)))))
     (dap-utils--get-extension url dest)))
 
+(defun dap-utils-vscode-get-installed-extension-version (path)
+  "Check the version of the vscode extension installed in PATH.
+Returns nil if the extension is not installed."
+  (require 'xml)
+  (require 'dom)
+  (let* ((extension-manifest (f-join path "extension.vsixmanifest")))
+    (if (f-exists? extension-manifest)
```
yyoncho, line 94: nit: when else clause is nil then you should use `when`.

abougouffa:
@yyoncho Thank you for the quick reply, fixed!

yyoncho:
> @yyoncho Thank you for the quick reply, fixed!

can you fix the other comment as well?

abougouffa:
@yyoncho Which comment?

yyoncho (approved):

yyoncho:
Thank you!

PR #623, opened 2022-05-13
Don't make setExceptionBreakpoints request if filters would be empty

themkat:
A lot of smaller debug adapters around the web does not support exception breakpoints, and a few I've tried even make the setExceptionBreakpoint request crash the adapter. (Example: some of these are assembly specific for 80s and 90s processors, like MOS). From what I can see in the specification of Debug Adapter Protocol, it is not necessary to send a setExceptionBreakpoints request if we do not have any. From the [specification](https://microsoft.github.io/debug-adapter-protocol/specification#Events_Initialized): "frontend sends a ‘setExceptionBreakpoints’ request if one or more ‘exceptionBreakpointFilters’ have been defined (or if ‘supportsConfigurationDoneRequest’ is not defined or false)"

While we could also make all adapters handle setExceptionBreakpoints, I think it makes more sense to just not send the request in dap-mode. This is because exception breakpoints only make sense for high level languages that support exceptions (Java, Kotlin etc.).

**EDIT:** The build issues seems to be because of issues with the setup-eask action and not because of my code.

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -1268,18 +1268,19 @@ DEBUG-SESSION is the active debug session."
 
 (defun dap--set-exception-breakpoints (debug-session callback)
   (-let [(&dap-session 'current-capabilities 'launch-args (&plist :type)) debug-session]
-    (dap--send-message
-     (dap--make-request "setExceptionBreakpoints"
-                        (list :filters
-                              (or (-some->> current-capabilities
-                                    (gethash "exceptionBreakpointFilters")
-                                    (-keep (-lambda ((&hash "default" "filter"))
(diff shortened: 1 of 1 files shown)


yyoncho (approved):
Thank you!

yyoncho:
@jcs090218 can you take a look at the CI? Seems like the setup is broken?

themkat:
I opened an issue about it https://github.com/emacs-eask/setup-eask/issues/22. Seems like @jcs090218  has fixed it now 🙂

jcs090218:
Sorry for the inconvenience! The issue was upgrading the dependency `toolchain`; I probably should not do that in the first place. 😅 Everything is good now!

PR #626, opened 2022-05-15
dap-dlv-go: fix an issue with test current function

kprav33n:
Due to incorrect usage of `-when-let`, "Test function" was not working
as expected. Fix the usage of `-when-let` to correctly update `:args`.


yyoncho:
cc @s-kostyaev

kprav33n:
Providing a link to show the example usage of `-when-let`.

https://github.com/magnars/dash.el#-when-let-var-val-rest-body

s-kostyaev:
Thank you!

PR #628, opened 2022-05-17
Fix for multiproject netcore solutions

### LINKED ISSUE #627: dap-netcore launch option causing error when running dotnet
When attempting to run the launch though dap debug there seems to be a problem with the way the dotnet run command for the dll is being built. For some reason there is a `-` after dotnet but even without that there needs to be a space after the dotnet and the location of the dll. Is this something that is a problem with the code or or the debug template?
```Could not execute because the specified command or file was not found.
Possible reasons for this include:
  * You misspelled a built-in dotnet command.
  * You intended to execute a .NET program, but dotnet-w:/personal/csharp/src/FluentUri/bin/Debug/FluentUri.dll does not exist.
  * You intended to run a global tool, but a dotnet-prefixed executable with this name could not be found on the PATH.```

### PR DESCRIPTION
**Pull request by s-kostyaev**
Fixes #627 

Instead of use `lsp-workspace-root` try to predict root of local project for find correct `.dll` location

### DIFF
--- a/dap-netcore.el
+++ b/dap-netcore.el
@@ -126,6 +126,34 @@ Will be set automatically in Emacs 27.1 or newer with libxml2 support."
         (error "Cannot start debugger configuration without netcoredbg")))
     dbg))
 
+(defun dap-netcore--locate-dominating-file-wildcard (file name)
+  "Starting at FILE, look up directory hierarchy for directory containing NAME.
+FILE can be a file or a directory.  If it's a file, its directory will
+serve as the starting point for searching the hierarchy of directories.
+Stop at the first parent directory containing a file NAME,
+and return the directory.  Return nil if not found.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: dap-netcore.el [hunk 1]
```diff
@@ -144,7 +176,7 @@ Will be set automatically in Emacs 27.1 or newer with libxml2 support."
 	    (f-join project-dir "bin" "Debug"
 		    (dom-text (dom-by-tag res 'TargetFramework))
 		    (dom-text (dom-by-tag res 'RuntimeIdentifier))
-		    (concat (f-base project-dir) ".dll")))))))
```
s-kostyaev, line 147: f-base works incorrectly with directories that contains dot in the name.

yyoncho (approved):

PR #631, opened 2022-06-09
test(makefile): add lint and checkdoc entries, remove unused variables

jcs090218:
We aren't using `DAP-GENERAL`, `DAP-CLIENTS` ... variables, I guess we can remove it?

Add `lint` and `checkdoc`, but left unused for now.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,11 +26,11 @@ jobs:
 
       - uses: actions/setup-node@v2
         with:
-          node-version: '14'
+          node-version: '16'
 
       - uses: emacs-eask/setup-eask@master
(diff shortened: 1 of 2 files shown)


s-kostyaev:
LGTM. Why new commands unused? Maybe we should put it to CI?

jcs090218:
`checkdoc` should be good. But I am not fully understand the code, so I doubt it would pass. 😅

Let's try it then. (see the next test)

jcs090218:
It won't pass due to multiple errors, see https://github.com/emacs-lsp/dap-mode/actions/runs/2466273845.

I have omitted `package-lint` for now.

PR #641, opened 2022-07-11
Fix dap--select-thread-id after dap--get-path-for-frame change

hmuniz:
After 5e449c8 dap--get-path-for-frame expects debug-session.


yyoncho (approved):

yyoncho:
Thank you!

PR #645, opened 2022-07-18
Add M-x customize parameter to set the screen position for the controls toolbar

genegoykhman:
I found the default position of the dap-ui controls toolbar interfered with the top line of my preferred layout, so this PR turns the position of the controls toolbar into a customizable variable `dap-ui-controls-screen-position`. The variable accepts, as a value, the positioning functions used by `posframe`.

For example:

`(setq dap-ui-controls-screen-position #'posframe-poshandler-frame-top-left-corner)`

will position the controls toolbar in the top-left corner of the screen (my preference).

The default value remains `#'posframe-poshandler-frame-top-center` so current users will not be affected.

### DIFF
--- a/dap-ui.el
+++ b/dap-ui.el
@@ -54,6 +54,16 @@ number - expand N levels."
                  (number :tag "Expand level"))
   :group 'dap-ui)
 
+(defcustom dap-ui-controls-screen-position #'posframe-poshandler-frame-top-center
+  "On-screen position of controls when they are visible."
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

yyoncho (approved):

codeasone:
Would be nice to see this merged.

yyoncho (approved):

PR #651, opened 2022-08-10
Update configuration.md

saltsucker:
Updated Python section to use and setup `debugpy`


vspinu:
Would be great if this is merged soonish. The ptsvd does no longer work for unclear reason. `debugpy` does work out of the box. Currently only way for new users to discover this is to search through github issues.

yyoncho (approved):

PR #662, opened 2022-08-19
Refactor debug provider names

### LINKED ISSUE #467: Inconsistency in the debug configuration type names between VSCode and dap-mode
The names used for some adapters for `dap-register-debug-provider` are not the same as the ones used by VSCode. This hinders the usage of the same `launch.json` file (with the configurations of these adapters) with both dap-mode and VSCode. Some of the ones that I have encountered are as follows:

| Adapter | VSCode | dap-mode|
| --- | :---: | :---: |
| CodeLLDB | `lldb`| `codelldb` |
| Native Debug(lldb) | `lldb-mi` | `lldb` |

I suggest that these should be refactored to match the ones used by VSCode.

MayushKumar:
fixes #467
Note: Will break any existing `launch.json` files and configurations using `codelldb` or `lldb` as type names.

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -2,6 +2,8 @@
 #+STARTUP: content
 
 * Changelog
+** 0.8
+   - [Breaking Change] Change debug provider names to match VS Code's naming: ~lldb~ to ~lldb-mi~ and ~codelldb~ to ~lldb~
 ** 0.7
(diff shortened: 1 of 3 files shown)


yyoncho:
please add changelog entry

yyoncho (approved):

yyoncho:
Thank you!

Issue #554, opened 2022-01-26
dap-node setup seems broken


barisere:
First, thank you to all the contributors for this extension. It has kept me from switching to VS Code.

I am unable to set up dap-node. It seems that the extension downloaded is broken. The logs from the _Messages_ buffer show that the ZIP file is empty, and I confirmed that in my file manager.

```txt
dap-node: ms-vscode.node-debug2 debug extension are not set. You can download it with M-x dap-node-setup
Contacting host: marketplace.visualstudio.com:443
uncompressing publicsuffix.txt.gz...done
Wrote /tmp/ext83ZK7l.zip
[/tmp/ext83ZK7l.zip]
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of /tmp/ext83ZK7l.zip or
        /tmp/ext83ZK7l.zip.zip, and cannot find /tmp/ext83ZK7l.zip.ZIP, period.
dap-node: Downloading done!
```

I'm be happy to provide more information if needed.

barisere:
@mihaiolteanu it is related. I fixed it by downloading the extension manually and extracting it in the destination directory.

abougouffa:
I got the same issue when running the `dap-gdb-lldb-setup` command, I tracked it down and it seems that the problem is the generated URL in `dap-utils-get-vscode-extension` (for me, it was: `https://marketplace.visualstudio.com/_apis/public/gallery/publishers/webfreak/vsextensions/debug/latest/vspackage`), the URL returns either this response:

```
{"$id":"1","innerException":null,"message":"Request was blocked due to exceeding usage of resource 'Count' in namespace 'AnonymousId'. For more information on why your request was blocked, see the topic \"Rate limits\" on the Microsoft Web site (https://go.microsoft.com/fwlink/?LinkId=823950).","typeName":"Microsoft.TeamFoundation.Framework.Server.RequestBlockedException, Microsoft.TeamFoundation.Framework.Server","typeKey":"RequestBlockedException","errorCode":0,"eventId":3000}
```

Or this one:

```
[...]

PR #700, opened 2022-12-27
Download node-debug2 from Open VSX registry

fredericgiquel:
Linked to: https://github.com/emacs-lsp/dap-mode/issues/554

Donwloading node-debug2 from VSCode Marketplace does not work. But the extension is available on Open VSX registry: https://open-vsx.org/extension/ms-vscode/node-debug2

This PR add support to download extensions from Open VSX registry and change `dap-node-setup` to use this feature.
Currently, direct download of the latest version of extension is not possible (https://github.com/eclipse/openvsx/issues/507). So it's down in 2 steps:
- query API (https://open-vsx.org/swagger-ui/index.html) to get metadata of extension and extract download URL
- download extension

PS: in addition to the technical difficulties related to VSCode Marketplace downloads, using Open VSX seems to be better from a legal point of view (according to https://www.eclipse.org/legal/open-vsx-registry-faq/).

### DIFF
--- a/dap-node.el
+++ b/dap-node.el
@@ -39,8 +39,8 @@
   :group 'dap-node
   :type '(repeat string))
 
-(dap-utils-vscode-setup-function "dap-node" "ms-vscode" "node-debug2"
-                                 dap-node-debug-path)
+(dap-utils-openvsx-setup-function "dap-node" "ms-vscode" "node-debug2"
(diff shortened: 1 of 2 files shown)


yyoncho:
Thank you for taking care of this. I will close https://github.com/emacs-lsp/dap-mode/issues/554

PR #701, opened 2022-12-28
Improve json-utils-sanitize-json

fredericgiquel:
Trying to debug a project with this [launch.json](https://github.com/ansible/ansible-language-server/blob/main/.vscode/launch.json) failed with a "JSON Readtable error".

I investigated the error and found that `json-utils-sanitize-json` does not remove the trailing comma on line 20 because the regexp does not correctly handle the case where there is only comments between comma and "]".

Rather than trying to add this case to the regexp, this PR splits the regexp in two and changes `json-utils-sanitize-json`  to run in 2 passes:
- first, remove comments
- second, remove trailing commas

This should make `json-utils-sanitize-json` more robust against this kind of cases.

### DIFF
--- a/dap-utils.el
+++ b/dap-utils.el
@@ -204,19 +204,30 @@ Replacement starts at point, and strings before it are ignored,
 so you may want to move point to `point-min' with `goto-char'
 first. This function moves `point'. Both // and /**/ comments are
 supported."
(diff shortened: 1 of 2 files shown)


yyoncho:
Thank you!

PR #702, opened 2022-12-31
Bump versions for gdb-lldb, codelldb, and cpptools

abougouffa:


jcs090218 (approved):

yyoncho:
Thank you. If you want to spent more time on this, you may check if we can move them to openvix just like in https://github.com/emacs-lsp/dap-mode/pull/700

abougouffa:
Thanks for the feedback, I will take a look at it when I get some free time.

PR #704, opened 2023-01-06
dap-dlv-go: add ability to test single subtest

### LINKED ISSUE #692: dap-dlv-go: Support debugging specific sub tests in Go
Currently we could debug a single test (dap-mode figure out test function name via `dap-dlv-go--extract-current--method-or-function-name` [here](https://github.com/emacs-lsp/dap-mode/blob/master/dap-dlv-go.el#L111) and pass it as argument to `-test.run` [here](https://github.com/emacs-lsp/dap-mode/blob/master/dap-dlv-go.el#L91).)

Would be nice to have some ways to debug Go sub tests instead of running a full test. 

Current workaround is to comment every other sub tests. But doesn't scale well for Go test with more than modest amount of sub tests.

s-kostyaev:
Fixes #692

### DIFF
--- a/dap-dlv-go.el
+++ b/dap-dlv-go.el
@@ -89,6 +89,13 @@
   (when (string= (plist-get conf :name) "Test function")
     (-when-let (name (dap-dlv-go--extract-current--method-or-function-name t))
       (dap--put-if-absent conf :args (list (format "-test.run=^%s$" name)))))
+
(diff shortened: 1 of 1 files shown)


yyoncho:
cc @kavirajk

s-kostyaev:
Should work for most cases of using table tests with gotests. For better results syntax + semantic parsing should be used.

PR #724, opened 2023-02-28
dap-python, Flask support

DJManas:
Now it should be all ok. Only thing left is what with the deprecated ptvsd, should I implement it also?

### DIFF
--- a/dap-python.el
+++ b/dap-python.el
@@ -221,10 +221,22 @@ strings, for the sake of launch.json feature parity."
        ;; them to unspecified instead. Some templates in this file set such
        ;; properties (e.g. :module) to nil instead of leaving them undefined. To
        ;; support them, sanitize CONF before passing it on.
-       (when program
-         (if module
-             (push program python-args)
-           (plist-put conf :program program)))
+       (if (string-equal module "flask")
+           ;; For flask module we need to strip filename from launch parameters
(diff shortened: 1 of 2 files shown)


codingquark:
I tried the changes PR makes in my emacs, with the following `.vscode/launch.json`:

```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Python App",
            "type": "python",
            "request": "launch",
            "module": "flask",
            "env": {
		"FLASK_APP": "app.py",
		"FLASK_DEBUG": "0"
            },
            "args": [
		"run",
		"--no-debugger",
		"--no-reload"
            ],
            "jinja": false,
            "justMyCode": true
        }
    ]
}

```

I use `direnv`, and had to override dap-mode function like so:
```elisp
(defun dap-python--pyenv-executable-find (command)
    (executable-find "python"))
```

Full config:
```elisp
(use-package dap-mode
  :after lsp-mode
  :commands dap-debug
  :custom
  (setq dap-python-debugger 'debugpy)
  (setq dap-python-terminal "vterm")
  :config
  (dap-auto-configure-mode)
  (require 'dap-node)
  (require 'dap-python)
  ;; no support for direnv, the function uses pyenv’s modules if pyenv
  ;; is found in path
  (defun dap-python--pyenv-executable-find (command)
    (executable-find "python")))
```

Pleased to report that things work well!

PR #736, opened 2023-05-03
MS `vscode-js-debug' Installation and Integration

jeff-phil:
Add latest vscode-js-debug adaptor, update installation routines so vscode adaptors on Github can be easily installed.  In order to make this work, updates for the latest DAP Spec for reverse requests.  A few other minor updates.

Please let me know if any questions at all.

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -5,6 +5,8 @@
 ** 0.8
    - [Breaking Change] Change debug provider names to match VS Code's naming: ~lldb~ to ~lldb-mi~ and ~codelldb~ to ~lldb~
    - Added ~dap-gdscript~
+   - Added ~vscode-js-debug~ latest Node.js, Chrome, Edge debuggers, along with latest DAP Spec reverse requests.
+   - Updates for extension installations
 ** 0.7
    - [Breaking change] For ~dap-lldb.el~, change ~type~ to ~lldb-vscode~.
 ** 0.5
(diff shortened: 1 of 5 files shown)


yyoncho:
There are some bits that should go in, and there are some pieces better to have from the other implementation. Let me think about it.

jeff-phil:
Sounds great.  Definitely try it out, and let me know if have any questions about what I've done, or want me to make changes, or anything else.

Thank you for taking the time to look at it!

jeff-phil:
Hi @yyoncho - Just checking if need anything from me on this PR.  Let me know if you do.

jeff-phil:
Resync'd with master.

Then made a update for `dap-js-debug.el` to work with `dap-debug-last` command; since the skeleton child debug configuration was being added to `dap--debug-configuration` variable after starting the child debug.  Needed to be popped off the list so that the parent debug conf is the last dap-debug.

Probably should determine a more upstream option for not adding the child confs automatically when starting debug, if the reverse request child debug sessions becomes more common.  For example, should there be logic in `dap-start-debugging-noexpand` to not add when `type` is missing - which is the case according to the spec.  E.g.:

```
(unless type
    (push (cons session-name launch-args) dap--debug-configuration))
```

What I did for now was look if `__pendingTargetId` which is from vscode-js-debug adapter in the last configuration and is the child.  That should be future proof regardless of how main `dap-mode.el` wants to handle.

Please, as always, let me know if have questions or feedback!  Thanks so much!

PR #739, opened 2023-05-10
Extract environment variables from launch.json

tinloaf:
The launch.json format and the dap-mode format for specifying environment variables is not the same, we need to do some conversion. First, launch.json uses the key 'environment', dap-mode uses 'environment-variables'. Second, launch.json uses a format like {'name': 'foo', 'value': 'bar'} to specify a single variable, while dap-mode uses a ('foo' . 'bar') cons cell.


yyoncho (approved):

yyoncho (approved):

yyoncho:
Thank you!

PR #748, opened 2023-08-09
Add Kotlin documentation to configuration guide

themkat:
dap-mode with Kotlin is sadly not as simple to setup as the Java one, nor is it as feature complete. From what I heard, many people are missing some documentation on how to setup, how it works etc. This is what I have done here. Some info on useful things like running JUnit tests is also covered. This is not really covered that many places, so I think it would be super useful to have the information in a place like this! 🙂 Might also get more interest in debugging and running Kotlin code in Emacs 🙂 (to be honest, I rarely use debuggers, but find it super nice to be able to run tests like this inside Emacs!)

A little bit unsure on the formatting, but tried to copy the style of what I could find...

### DIFF
--- a/docs/page/configuration.md
+++ b/docs/page/configuration.md
@@ -73,6 +73,52 @@ settings.
                                        :env '(("DEV" . "1"))))
     ```
 
+## Kotlin
+
+1.  Installation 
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

yyoncho (approved):

yyoncho:
Thank you!

PR #768, opened 2023-12-15
fix: get list of buffers (not files) with breakpoints

sfavazza:
This is a fix for a bug I erroneously introduces in PR #754 . 

The function `dap--buffers-w-breakpoints` was returning a list of files names instead of a list of buffers. This in turn broke the `dap--switch-to-session`.

With this fix, the list of buffers is inferred by the first marker of the first breakpoint found for each file.

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -861,8 +861,10 @@ will be reversed."
 
 (defun dap--buffers-w-breakpoints ()
   "Get only the buffers featuring at least one breakpoint"
-  ;; get the list from the keys of the breakpoint hash-table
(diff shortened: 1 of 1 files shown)


sfavazza:
@yyoncho is it critical that only the `CI / test (windows-latest, snapshot, true)` check does not pass?

jcs090218:
> is it critical that only the CI / test (windows-latest, snapshot, true) check does not pass?

There is a bug with that emacs snapshot. We can ignore that for now.

jcs090218 (approved):
LGTM.

PR #785, opened 2024-04-06
Add customization option to control expand depth for dap-ui-sessions

bastianbeischer:
This defcustom controls the depth to which dap-ui-sessions expands by default


jcs090218 (approved):
LGTM.

PR #802, opened 2024-08-03
dap-hydra: make dap-debug-edit-template exit

daanturo:
Stop the hydra to edit text normally.

I think this feels more "natural". `dap-debug-edit-template` brings up another buffer for editing, exit the hydra to free keys for that purpose.


jcs090218 (approved):

jcs090218:
Thank you!

PR #805, opened 2024-08-26
Consolidate and update native debug instructions

perrys:
Hi. I'm new to dap-mode, and I found the instructions for the native debuggers a little sparse and out-of-date in places. After spending some time figuring out how they work, I've updated the documentation to (hopefully) make things clearer and explain the configuration options in more detail. HTH, Stu Perry.


davidhaley:
Leaving a message here to say that you documentation helped me. Thank you.

jcs090218 (approved):

PR #807, opened 2024-09-06
fix: check if body is null

oxidase:
I got no body in a response with `success` status
```
Sending:
{
  "command": "source",
  "arguments": {
    "sourceReference": 0
  },
  "type": "request",
  "seq": 6
}
Received:
{
  "seq": 17,
  "type": "response",
  "request_seq": 6,
  "command": "source",
  "success": true
}
#s(hash-table size 65 test equal rehash-size 1.5 rehash-threshold 0.8125 data (seq 17 type response request_seq 6 command source success t))
```
and dap-mode failed with 
```
error in process filter: Wrong type argument: hash-table-p, nil [2 times]
```

This is a quick fix to check if body is non-nil


jcs090218:
Thank you!

PR #824, opened 2025-01-30
Re-enable dapui.el so I can see loaded sources

timfel:
I am debugging some embedded GraalPy code things with dynamically generated Python sources that I can attach to via DAP but there are no real files to open.

I simply followed https://github.com/Alexander-Miller/treemacs/blob/master/Extensions.org to update the previous code to the new APIs


timfel:
I suppose the name `dapui` for the module is quite misleading and should be changed, or maybe the entire code should just be moved into `dap-ui.el`?

jcs090218 (approved):

jcs090218:
Thank you!

PR #828, opened 2025-02-16
Interactive netcore test command

sjlwa:
Hello. This adds functions to use dap-netcore with tests.
`dap-netcore-debug-test` automatically runs the tests and attach the debugger.


jcs090218 (approved):

jcs090218:
Thank you!

PR #837, opened 2025-03-10
fix: fixes error when a utility frame is the current frame

ryanobjc:
On Mac OS X, small frames that don't allow user focus may be selected when the process filter is processed.  By walking up the parent-frame list, we can get to the frame that the user is actually able to use, and thus the call to (get-mru-window) does not return nil and the whole process filter aborts.

Should fix bug #828


SerialVelocity:
@jcs090218 Sorry for the direct ping, but would you be able to take a look at this?

It also fixes dap-java for me locally (also using Mac OS X, Emacs 29)

csalamea-itti (approved):

jcs090218 (approved):

jcs090218:
Sorry for the late reply! Merged!

Issue #126, opened 2019-09-20
Problem manually setting dap-python-executable


rr326:
I have been struggling getting dap to work with my python virtual environment. As a workaround, @yyoncho suggested manually setting `dap-python-executable` to the full path of python. 

After doing that, I get:
```
nil -m ptvsd --wait --host localhost --port 32000 /Users/rrosen/dev/kachess/kachess/kdata.py 
zsh:1: command not found: nil
```

@yyoncho found the problem. The workaround is: you may fix it by going to dap-python.el and changing line 70 to `(python-executable dap-python-executable)`

He thinks: The issue has been introduced probably with emacs-lsp/dap-mode#105 @mpanarin

(BTW - The workaround works.)

rr326:
Additional point - before setting the full path, dap-python was using the python of my .zshenv, even after I manually did `M-x pyvenv-activate` which properly sets the `VIRTUAL_ENV` and `PATH`. Perhaps virtualenv problem is related. (And note - when I view `process-environment` it properly shows `PATH` and `VIRTUAL_ENV`, as you'd expect.)

yyoncho:
@rr326 can you provide more details on your python configuration.

rr326:
I'm on OSX 10.14.6, which might matter.

mpanarin:
@rr326  My comment may be dumb, but just in case:
Did you restart lsp workspace after activating venv?
Because, afair, lsp doesn't update itself on venv activation and this may be an issue.

fangherk:
I'm getting the same error as well. I use both `pyenv` and `pyvenv`, but I'm mainly working in a `pyvenv` environment. It seems that https://github.com/emacs-lsp/dap-mode/blob/master/dap-python.el#L49 is defaulting to finding pyenv by default if it exists, but a prompt might be more useful here.

yyoncho:
> but a prompt might be more useful here.

Fallback to interactive python lookup sounds good in case user have some misconfiguration and dap-python is unable to find the correct binary.

honnix:
I'm experiencing the same issue and my workaround is to override `dap-python--pyenv-executable-find` as:
```elisp
  (defun dap-python--pyenv-executable-find (command)
    (executable-find command))
```

Since `pyvenv` sets up proper environment variables, `executable-find` works just fine.

Another thing to note is, without activating any virtualenv via `pyvenv`, `(executable-find "python")` results a proper shim, for example in my case `$HOME/.pyenv/shims/python`.

PR #839, opened 2025-03-27
Use pyenv to find executable only if pyenv is installed and a local shim exists for the cwd

gamboz:
I have a problem similar to that discussed in #126 i.e. I have `pyenv` installed, but I'm working on `pyvenv` environment with no `pyenv` "local" application-specific Python version for the current project / cwd.

Here is a "this-works-for-me" change that lets dap-mode use `pyenv` only if `pyenv local` returns something different than "pyenv: no local version configured for this directory"

Unfortunately I'm not expert enough to be sure that what I'm proposing here is always a good solution, so please feel very free to close this PR :slightly_smiling_face:

### DIFF
--- a/dap-python.el
+++ b/dap-python.el
@@ -51,8 +51,13 @@ you are debugging."
   "Find executable COMMAND, taking pyenv shims into account.
 If the executable is a system executable and not in the same path
 as the pyenv version then also return nil. This works around
-https://github.com/pyenv/pyenv-which-ext."
-  (if (executable-find "pyenv")
+https://github.com/pyenv/pyenv-which-ext.
+
+If pyenv is not installed or if pyenv has no \"local\" version defined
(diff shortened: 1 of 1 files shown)


dap-python.el
```diff
@@ -50,9 +50,15 @@ you are debugging."
...
-  (if (executable-find "pyenv")
+as the pyenv version then also return nil.  This works around
+https://github.com/pyenv/pyenv-which-ext.
+
+If pyenv is not installed or if pyenv has no \"local\" version defined
+for the cwd, then fallback to `(executable-find)`, which see."
+  (if (and
+       (executable-find "pyenv")
+       (not (string-match "no local" (shell-command-to-string "pyenv local")))
+       )
```
jcs090218: style issues:

- Don't leave bracket on it's own.
- Move line 59 after `(and` and re-indent the region.

Issue #726, opened 2023-03-13
dap-compilation  seems fail to execute


zhyzky:
When using dap-debug, with template containning dap-compilation, first execute
```
        (dap-debug-run-task `(:cwd ,(or (plist-get launch-args :dap-compilation-dir)
                                         (lsp-workspace-root)
                                         default-directory)
                               :command ,dap-compilation
                               :label ,(truncate-string-to-width dap-compilation 20)) cb)
```
then in function call to dap-debug-run-task, there are some codes
```
        let* ((task (if (listp tasks)
                     (car tasks)
                   tasks)))
```
with task value becomes :cwd, which seems not the expected result, and breaks the dap-compilation run.

zhyzky:
I got a workaround, change the first code block like this
```
 (dap-debug-run-task `((:cwd ,(or (plist-get launch-args :dap-compilation-dir)  ;; <--  `( to `((
                                         (lsp-workspace-root)
                                         default-directory)
                               :command ,dap-compilation
                               :label ,(truncate-string-to-width dap-compilation 20))) cb)
```
Seems to close this issue.
Hope if there's better sollution.

PR #840, opened 2025-03-31
Fix 726 improve dap compilation

Ndot:
Fixes issue #726 

There were some bugs in the way the TASKS were being check. It also set the hook for `compilation-start` after the functions was already called and banked on the fact that the command took long enough for us to set the hook, which seem like a strange thing to do.

Also added two options: 
- `dap-debug-compilation-keep-buffer` to allow deleting the buffer on success.
- `dap-debug-compilation-same-buffer` to allow the commands to run on the same buffer without clearing the buffer.

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -1838,6 +1838,15 @@ it would not make sense to kill the buffer and leave the window open."
   :type 'boolean
   :group 'dap-mode)
 
+(defcustom dap-debug-compilation-same-buffer nil
+  "If non-nil commands run by `:dap-compilation' will run in the same buffer
+without clearing the buffer.
+
+This only works if the buffers share the same name.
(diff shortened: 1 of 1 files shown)


dap-mode.el
```diff
@@ -1820,41 +1820,82 @@ before starting the debug process."
...
+(defcustom dap-debug-compilation-same-buffer nil
+  "If non-nil commands run by `:dap-compilation' will run in the same buffer
+without clearing the buffer.
+
+This only works if the buffers share the same name.
+This variable is passed to `compilation-start' as the CONTINUE argument."
+  :type 'boolean
+  :group 'dap-mode)
+
+(defun dap-debug-run-task--cf (buf status)
```
jcs090218, line 1850: Can you add docstring to this function? Thanks!

PR #848, opened 2025-10-03
Fix byte compiler and native compiler warnings

pierre-rouleau:
Fix a defcustom and function not known to be defined warnings.

### DIFF
--- a/dap-launch.el
+++ b/dap-launch.el
@@ -37,42 +37,44 @@ Yields nil if it cannot be found or there is no project."
 (defun dap-launch-get-launch-json ()
   "Parse the project's launch.json as json data and return the result."
   (require 'dap-utils)
-  (when-let ((launch-json (dap-launch-find-launch-json))
-			 (json-object-type 'plist)
-			 ;; Use 'vector instead of 'list. With 'list for array type,
-			 ;; json-encode-list interpreted a list with one plist element as
-			 ;; an alist. Using 'list, it turned the following value of
-			 ;; pathMappings:
-			 ;;
-			 ;;     "pathMappings": [
-			 ;;         {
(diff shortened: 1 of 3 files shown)


pierre-rouleau:
All CI failures are failing on timeout...  
Running a eask make on my system succeeds.

That's the second eask-based CI controlled project where test fail because of timeouts. Is this an issue with the Github infrastructure?

jcs090218:
> That's the second eask-based CI controlled project where test fail because of timeouts. Is this an issue with the Github infrastructure?

The failure occurred because of the slow internet speed from the GNU main FTP server during the Emacs installation. If you visit their FTP site, you may notice the same slowness. 🤔

pierre-rouleau:
Hi @jcs090218,  I did notice the speed of several GNU site going down in the last months.   It seems slower than it was before. I wonder if this is due to AI scraper traffic or changes in the routing rules affected by politics.

pierre-rouleau:
Hi @jcs090218 , on the topic of CI and Emacs installation, would it not be possible to pre-package images with Emacs and necessary pre-packaged stored in a location that has higher access speed to alleviate the problem?

jcs090218:
> Hi @jcs090218, I did notice the speed of several GNU site going down in the last months. It seems slower than it was before. I wonder if this is due to AI scraper traffic or changes in the routing rules affected by politics.

No, I’m not sure if it’s because of AI scraping the entire internet, but that might be the case. 🤔

> Hi @jcs090218 , on the topic of CI and Emacs installation, would it not be possible to pre-packaged images with Emacs and necessary tools stored in a location that has higher access speed to alleviate the problem?

Yeah, that would be the ideal solution. This repo is a good starting point: https://github.com/Silex/docker-emacs. Unfortunately, it doesn’t support Windows images.

PR #851, opened 2026-01-23
Check supportsConfigurationDone capability before running configurationDone request

themkat:
Experienced a server that didn't support `configurationDone` type requests. The server was sending the correct information in the initialize response, but dap-mode still sends the request. That caused issues in the server without giving a response. dap-mode then was stuck in a non-initialized state. 

This fix simply checks if the server supports `configurationDone` by checking the capabilities. If it is supported, we just send it like normally. If not, we just run the success callback to start dap. In the cases I've tried, this resolves all issues.

### DIFF
--- a/dap-mode.el
+++ b/dap-mode.el
@@ -1283,13 +1283,16 @@ ADAPTER-ID the id of the adapter."
 
 (defun dap--send-configuration-done (debug-session)
   "Send `configurationDone' message for DEBUG-SESSION."
-  (dap--send-message (dap--make-request "configurationDone")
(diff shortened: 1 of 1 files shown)


themkat:
@jcs090218 , any chance you have time to take a quick look in the near future? :slightly_smiling_face:

jcs090218 (approved):

jcs090218:
Sorry for the late reply. I'll merge this now since it make sense to me! :)

jcs090218:
Thank you for fixing this!

PR #854, opened 2026-05-05
chore: Add fontification for code fetched by sourceReference when navigating the stack

Konubinix:
The code tries hard to guess the correct type, using either the path or the name of the source.

It also use buffer-read-only to make clearer that this is only a view of something remote and not a file to be edited.

Note: I discovered that something similar is done in dap-ui and initially tried to refactor them, but I failed. dap-ui appears to expect the file to eventually go to a file, while dap-mode does not.


Konubinix:
Hi, anything I can do to help review that PR ?

Konubinix:
Hi

PR #855, opened 2026-05-23
Drop extra progn

gamboz:
There is a `(progn (let ...))` fragment where the `(progn)` is probably useless.

Here I propose to remove it.

The diff could show a lot of changes because of the white-space changes of re-indenting the `let` after removing the `progn`.

AFAIK, this is mostly aesthetic, feel free to disregard :slightly_smiling_face:


jcs090218 (approved):

jcs090218:
Thank you!

Issue #808, opened 2024-09-07
Wrong type argument stringp nil with dap-lldb on Mac M1


lanceberge:
My configuration:
```emacs-lisp
(use-package cc-mode
  :ensure nil
  :mode ("\\.cu\\'" . c-mode)
  :init
  (require 'dap-lldb)
  :hook
  (c++-mode . +cpp-mode)
  (c++-ts-mode . +cpp-mode)
  :custom
  (c-basic-offset 4)
  (c-default-style "linux")
  (dap-lldb-debug-program "/opt/homebrew/opt/llvm/bin/lldb-dap") ; this executable does exist
  :config
  (defun +cpp-mode ()
    (setq-local tab-width 4
                dap-debug-template-configurations nil)

    (dap-register-debug-template
     "lldb"
     (list :type "lldb-vscode"
           :request "launch"
           :name "lldb"
           :MIMode "lldb-dap"
           :program "${workspaceFolder}/average_contiguous_subarray" ; the file I'm trying to debug
           :cwd "${workspaceFolder}"))))
```

I do dap-debug, then get:
```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  #<subr make-process>(:name "lldb" :connection-type pipe :coding no-conversion :command "/opt/homebrew/opt/llvm/bin/lldb-dap" :stderr "*lldb stderr*" :noquery t)
  make-process@with-editor-process-filter(#<subr make-process> :name "lldb" :connection-type pipe :coding no-conversion :command "/opt/homebrew/opt/llvm/bin/lldb-dap" :stderr "*lldb stderr*" :noquery t)
  apply(make-process@with-editor-process-filter #<subr make-process> (:name "lldb" :connection-type pipe :coding no-conversion :command "/opt/homebrew/opt/llvm/bin/lldb-dap" :stderr "*lldb stderr*" :noquery t))
  make-process(:name "lldb" :connection-type pipe :coding no-conversion :command "/opt/homebrew/opt/llvm/bin/lldb-dap" :stderr "*lldb stderr*" :noquery t)
  apply(make-process (:name "lldb" :connection-type pipe :coding no-conversion :command "/opt/homebrew/opt/llvm/bin/lldb-dap" :stderr "*lldb stderr*" :noquery t))
[...]

PR #856, opened 2026-06-12
Probe lldb-dap/lldb-vscode on PATH for dap-lldb-debug-program default

jlipworth:
LLVM 18 renamed `lldb-vscode` to `lldb-dap`, and the `llvm-org.lldb-vscode-0.1.0` VSCode extension that the current default points at no longer exists, so out of the box `dap-lldb` fails with a missing-program error on any current system. This appears to be the root cause behind #808 and #800 (and spacemacs's syl20bnr/spacemacs#16289).

This changes the `dap-lldb-debug-program` default to probe `executable-find` for `lldb-dap`, then `lldb-vscode`, before falling back to the historical extension path, and updates the Commentary link to the adapter's current home in llvm-project (the old llvm-mirror tree is gone).

Verified end to end on macOS (Homebrew LLVM 20, Emacs 30.2): with the new default, `dap-debug` with the `lldb-vscode` provider launches `lldb-dap`, stops at a source breakpoint in a C++ program, and runs to termination with program output captured.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -5,6 +5,7 @@
 ** Unreleased 0.9
    - Added ~dap-gdb~
    - Added ~dap-julia~
+   - ~dap-lldb-debug-program~ now defaults to ~lldb-dap~ (or the older ~lldb-vscode~) found on ~PATH~, falling back to the historical VSCode extension location.
 ** 0.8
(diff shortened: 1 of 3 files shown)


jcs090218 (approved):

jcs090218 (approved):

jcs090218:
Thank you!

